PR #2, opened 2016-12-29
Add CMake & Linux x86_64 fixes

z33ky:
I didn't touch the Code::Blocks project file just in case, but you should be able to remove it in favor of this CMake file.
To get a Code::Blocks project out of it, run `cmake -G "CodeBlocks - MinGW Makefiles"` or use `cmake-gui`. You should use a dedicated build-directory.

I use stricter warning flags in my projects (`-Wall` & co), so I've included them here as well.
The Windows paths, including MSVC, is not tested... But on Linux x86_64 I get a binary that runs, although it's using the stubbed vectors maths stuff.

You can take a look at my master branch, there's other changes that you can pull if you want.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,85 @@
+cmake_minimum_required(VERSION 3.1)
+
+execute_process(COMMAND cmake -E echo *
+	OUTPUT_FILE ".gitignore")
+
+project(sh3rd CXX)
+
+if(WIN32)
+	enable_language(RC)
+endif()
+
+if("${CMAKE_SIZEOF_VOID_P}" EQUAL 8)
+	set(BINARY_RESOURCE_FILE "resource64.rc")
+	add_definitions(-DSH3_64)
+elseif("${CMAKE_SIZEOF_VOID_P}" EQUAL 4)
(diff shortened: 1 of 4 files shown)


Quaker762:
Hi, 

I'll keep both for now, as I'm not too sure about CMake (I've used it before, but don't know it too well).
Would it be a good idea to split different project types into a 'build' folder eventually (so even more systems can be targeted)?

Quaker762:
Oh, and by the way, feel free to make more pull requests, I had a look and it seems you've fixed some of the spaghetti I wrote haha.

Quaker762:
I'm having an issue with CMake, it seems it can't find GLEW (I'm using it locally, not sure if you are on your system). Do you have any idea as to why this is (build systems aren't my thing haha)

z33ky:
Installing GLEW in the standard paths should work (see http://glew.sourceforge.net/install.html).
Alternatively you can set additional paths to look in by setting `CMAKE_PREFIX_PATH` or `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH`.
You can also directly set `GLEW_INCLUDE_DIR` and `GLEW_LIBRARY` to the correct paths.

PR #3, opened 2016-12-30
Remove 3rd party library stuff

z33ky:
It's not customary to have this stuff in the repository.
The CMake build system should take care of the dependencies and allows for either system-installed versions to be used or user-installed versions (non-standard paths, including putting it in the same directory as the repository).
If bundeled dependencies are desired, a better approach would be to integrate it as an external project in the build system, which will also automatically build it for the target system.


huming2207:
Yep I agree with you. We should use the library from OSes' own paths, as we need to port this project to various platforms.

PR #7, opened 2016-12-30
Fixes and improvements to Log()

z33ky:
I've seen that `Log(LOG_FATAL, ...)` is almost always followed by `messagebox()` and always with `exit()`. In another commit, this could be put in a function like `FatalError()`, which does all three things.


Quaker762:
Thanks, I probably should've done that, slipped my mind haha.

PR #8, opened 2016-12-30
Add sh3_arc_file

z33ky:
I think this makes a nicer API rather than directly using zlib.

### DIFF
--- a/include/SH3/arc/sh3_arc_file.hpp
+++ b/include/SH3/arc/sh3_arc_file.hpp
@@ -1,0 +1,59 @@
+/*++
+
+Copyright (c) 2016  Palm Studios
+
+Module Name:
+        sh3_arc_file.hpp
+
+Abstract:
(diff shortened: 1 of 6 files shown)


Quaker762:
Looks good. I was going to do something like this, I honestly never expected anyone to fork the repo haha, so it was pretty dirty code.

z33ky:
Someone posted a like at the linux_gaming reddit, where I'm lurking regularly. Porting stuff to Linux spikes my interests, so I attempted that and just glossed over some other stuff while I have plenty of free time between the years.
I can't guarantee that I'll stay a contributor for a longer time, but we'll see what happens.

Quaker762:
Cheers mate, any contribution is good. I'd definitely like to get more people on this and actually get it rolling. SH3  (and SH2 for that matter) needs a modern update (and for linux).

P.S I just saw the link, didn't think anyone would share it, holy shit. The pressure is on.

PR #10, opened 2016-12-31
Created a 'die( )' function (to reduce code volume)

Quaker762:
Also removed a lot of messagebox() code bloat.

### DIFF
--- a/include/SH3/stdtype.hpp
+++ b/include/SH3/stdtype.hpp
@@ -86,6 +86,9 @@ static inline void messagebox(const char* title, const char* str, ...)
 #define SHFALSE 0
 #define SHTRUE  1
 
+#define SH_OK   0
+#define SH_BAD  -1
+
 /**++
     These allow us to check more specifically for a bad allocation across multiple systems
     though the compiler/stdlib really should take care of it for us
@@ -101,7 +104,7 @@ static inline void messagebox(const char* title, const char* str, ...)
     #define BADALLOC (void*)0; // Not sure what Linux/OSX define this as...
 #endif
 
-typedef int SHSTATUS;
+typedef int SHSTATUS; // Return type for functions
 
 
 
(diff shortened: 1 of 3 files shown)


include/SH3/stdtype.hpp
```diff
@@ -115,8 +118,37 @@ typedef int SHSTATUS;
...
+        the error log, log.txt)
+
+Arguments:
+        str - String to display to the user
+
+Return Type:
+        None
+
+--*/
+static inline void die(const char* str, ...)
```
z33ky, line 141: This looks like a relatively large function that should not be inlined.

include/SH3/stdtype.hpp
```diff
@@ -115,8 +118,37 @@ typedef int SHSTATUS;
...
+{
+    va_list args;
+    char    buff[4096];
+
+    va_start(args, str);
+    vsnprintf(buff, sizeof(buff), str, args);
+    Log(LOG_FATAL, buff);
+    messagebox("Fatal Error", buff);
+    va_end(args);
+    exit(SH_BAD);
```
z33ky, line 151: Neither Windows nor Linux have good support for negative exit codes.

include/SH3/stdtype.hpp
```diff
@@ -86,6 +86,9 @@ static inline void messagebox(const char* title, const char* str, ...)
 #define SHFALSE 0
 #define SHTRUE  1
 
+#define SH_OK   0
```
z33ky, line 89: Should `SH_{OK,BAD}` not follow the same convention as `SH{TRUE,FALSE}` (either with or without underscore)?

Also I think an enum would be good instead of macros.

z33ky (commented):
Not really part of this change, but I'm not convinced that the `SHSTATUS` typedef is useful, which would include `SH_{OK,BAD}`.  
I'm also doubtful that `SH{TRUE,FALSE}` is useful when C++ has `bool`.

Quaker762:
Sorry I wrote this right before a New Years party so I was in a bit of a rush. I'm more of a C programmer, so I'm used to having to typedef'ing stuff and using macros. By all means feel free to change it, I just like the extra level of abstraction.

z33ky:
That explains the C-style string and memory handling :P
Props for choosing C++11, it has received some nice additions there. The 14 and upcoming 17 standard include more nice things, but obviously require a newer compiler.

I'm all in favour of abstractions if there's actually something to abstract over. Abstraction for abstraction's sake just obscures things IMO. I'll just change it and send more pull requests for review.

Quaker762:
Haha yeah, usually when I use C++ it's just "C with classes" (like it's the 90s or something). 
Yeah I totally agree, I feel like it's the best standard to use, as 14 and 17 are drifting more and more away from C and becoming quite cryptic, hence C++11 for everything hahaha. 

And sure thing, some of my programming style comes from reading over old code from Microsoft, where almost every type is an abstraction (which I assume is to avoid confusion in a larger codebase). Feel free to change anything you think is unnecessary.

z33ky:
I feel like C++14 is more an addendum to C++11, with things like `std::make_unique`, relaxed `constexpr`, lambda capture expressions, `{c,b}{begin,end}` all enhancing things that C++11 introduced.

As a side note, (Boost)[http://www.boost.org], as over-engineered as it is, does have some cool libraries. (Boost.Range)[http://www.boost.org/doc/libs/release/libs/range] is probably the one I use the most. (Boost.Filesystem)[http://www.boost.org/doc/libs/release/libs/filesystem] is also handy.

PR #17, opened 2017-01-01
File resturcture

z33ky:
On top of that I'd also be in favour of dropping the `sh3_` prefix for the types and replacing them with the namespace `sh3`. Likewise for `arc_`.
However, I'd only put the Silent Hill 3 specific parts in that namespace and I'm not confident enough to make this division.


Quaker762:
Wouldn't hand done assembly be faster, or is GCC smart enough to be able to compile it down with SSE in 5/6 lines like I managed to?

z33ky:
Whoops, wrong branch :)

PR #20, opened 2017-01-01
Use __attribute__((vector_size())) to leverage SSE

z33ky:
You can check that it does work by examining the disassembly. Set `CMAKE_CXX_FLAGS` to '-msse' to make sure it actually uses SSE. You can also use '-mavx' to make it use AVX instead. Likewise, '-march=pentium2' will spit out good old x87 instructions.
You will need the stuff in #19 to make setting `CMAKE_CXX_FLAGS` work.


z33ky:
To your question in #17, `clang -Os` can. `gcc -Os` output is larger, but I know not to trust my limited knowledge in CPU microarchitecture to say that smaller is strictly better.

Edit: At other optimization levels (1-3), clang also produces the small output.

Quaker762:
It seems the last PR changing the file names has created a merge conflict.

z33ky:
Rebased, no changes needed.

PR #21, opened 2017-01-01
File resturcture

z33ky:
On top of that I'd also be in favour of dropping the `sh3_` prefix for the types and replacing them with the namespace `sh3`. Likewise for `arc_`.
However, I'd only put the Silent Hill 3 specific parts in that namespace and I'm not confident enough to make this division.


Quaker762:
Sounds good, again the typing (sh3_, arc_) is due to my C style of programming. Namespaces sound much better.

Quaker762:
Shit, I thought comment and close would MERGE... Oops

PR #24, opened 2017-01-01
Use proper std:: prefixes

z33ky:
It working without it is a C relic, not guaranteed to work in C++.

Also includes some more headers in places where they were missing.

### DIFF
--- a/include/SH3/arc/file.hpp
+++ b/include/SH3/arc/file.hpp
@@ -23,6 +23,7 @@ Revision History:
 
 #include <zlib.h>
 
+#include <cstddef>
 #include <memory>
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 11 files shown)


Quaker762:
Is there any reason for this besides it being wrapped in the std namespace??

z33ky:
I don't understand the question. If it wasn't in the `std` namespace, this wouldn't work.
And it working without specifying the `std` namespace is an implementation detail that just happens to work with most C standard library types and functions and is thus a portability concern. This is like a `file.cpp` being able to use stuff declared in `a.hpp` although it doesn't directly `#include` it, but file `b.hpp`, which is `#include`d by `file.cpp`, does. If `a.hpp` is changed so that it doesn't `#include` `b.hpp` anymore, `file.cpp` will fail to build. Likewise if `b.hpp` is part of a library that is replaced with another drop-in implementation, or just a differently configured version of the same library, but does not `#include` `a.hpp` from `b.hpp`.
Side note: There are situations where you want to avoid specifying the concrete namespace: Look up "argument-dependant lookup" (ADT).

PR #25, opened 2017-01-02
sh3_arc fixes

z33ky:


Quaker762:
I'm only now seeing how much easier all of this makes life. My god, the C style syntax I was using is already starting to feel archaic (which it is haha)..

z33ky:
C still has it's place though: it doesn't hide a great deal. It's very explicit.
With C++ you can build beautiful abstractions (horrible ones as well...), but you can hide a great many of things, mostly implicit function calls (think destructors or overloading operators), but also other stuff (like virtual function lookups or hidden code bloat via templates).

People can learn to write good code in any (decent) language, but in C it will be more apparent if it is not, which is why Linus Torvalds, for instance, is not a fan of C++.

PR #30, opened 2017-01-02
Basic texture handling

Quaker762:
Implemented some basic texture handling, this relies on holding a reference to sh3_arc, so definitely needs to be refactored.

### DIFF
--- a/include/SH3/graphics/texture.hpp
+++ b/include/SH3/graphics/texture.hpp
@@ -34,22 +34,22 @@ namespace sh3_graphics
         std::uint8_t  unused1[4];                 /** There are a lot of unused DWORDs, I assume to align everything nicely */
         std::uint32_t batchHeaderSize;            /** Size of the first part of the whole header */
         std::uint32_t batchSize;                  /** = res * res * (bpp/8) * #tex + 128 * #tex */
-        std::uint8_t  unused2[4];
+        std::uint8_t  unused2[4];                 /** Unused **/
         std::uint32_t numBatchedTextures;         /** Number of textures in this texture file */
-        std::uint8_t  unused3[8];
+        std::uint8_t  unused3[8];                 /** Unused **/
         std::uint32_t texHeaderSegMarker;         /** Secondary texture marker. This signifies the start of the texture information header */
-        std::uint8_t  unused4[4];
+        std::uint8_t  unused4[4];                 /** Unused **/
         std::uint16_t texWidth;                   /** The width of this texture */
         std::uint16_t texHeight;                  /** The height of this texture */
         std::uint8_t  bpp;                        /** Number of bits per pixel of this texture. NOTE: 8bpp is believed to be paletted! */
         std::uint8_t  dataOffset;                 /** #bytes from texHeaderSize+16 to 128 (0 filled) */
-        std::uint8_t  unused5[4];
+        std::uint8_t  unused5[4];                 /** Unused **/
         std::uint32_t texSize;                    /** The size of this texture in bytes (w * h * [bpp/8]) */
         std::uint32_t texHeaderSize;              /** = texSize + texHeaderSize + 16 + endFilleSize */
-        std::uint8_t  unused6[4];
+        std::uint8_t  unused6[4];                 /** Unused **/
         std::uint32_t unknown1;                   /** Completely unknown, probably unimportant for now */
         std::uint32_t unknown2;                   /** Usually 1! */
-        std::uint32_t unused7[15];
+        std::uint32_t unused7[15];                /** Unused **/
     } sh3_texture_header;
 
     /**
@@ -57,7 +57,7 @@ namespace sh3_graphics
      * Describes a logical texture that can be bound to OpenGL
      *
      * Defines a few bits of data and some functions to load in a texture
-     * from a SILENT HILL 3 .arc section.
+     * from a SILENT HILL 3 .arc section.both batch and individual texture
      *
      * @note
      *
@@ -65,28 +65,28 @@ namespace sh3_graphics
     class texture
     {
     public:
-    enum class result
-    {
-        INVALID_CHUNK,
-        TEOF,
-
-    };
+        enum class load_result
+        {
+            INVALID_CHUNK,
+            TEOF,
+        };
 
         texture(){}
-        texture(std::string& filename, sh3_arc& arc){Load(filename, arc);}
+        texture(const std::string& filename, sh3_arc& arc){Load(filename, arc);}
         ~texture(){}
 
-    /**
-     * Load a texture and return a result to it
-     *
-     * @note
-     *
-     * @param filename - Virtual file path to file.
-     *
-     * @return \c result indicating whether the read failed or not
-     *
-     */
-        result Load(const std::string& filename, sh3_arc& arc);
+        /**
+         * Load a texture and return a result to it
+         *
+         * @note
+         *
+         * @param filename - Virtual file path to file.
+         * @param arc - Reference to sh3_arc (arc.arc)
+         *
+         * @return @ref load_result indicating whether the read failed or not
+         *
+         */
+        load_result Load(const std::string& filename, sh3_arc& arc);
 
     private:
         std::vector<std::uint8_t>       pixbuff;
--- a/include/SH3/graphics/texture.hpp
+++ b/include/SH3/graphics/texture.hpp
@@ -1,56 +1,89 @@
-/*++
-
-Copyright (c) 2016  Palm Studios and Mike M (@perdedork)
-
-Module Name:
-        sh3_texture.hpp
-
-Abstract:
-        Defines data structures to load an manipulate a texture
-
-Author:
-        Jesse Buhagiar
-
-Environment:
-
-Notes:
-        I'd like to thank Mike for all this, he put a lot of time into reverse engineering
-        all of the file types, including all of the textures, meshes and even the motion capture
-        skeletal animation that Konami captured. Thanks mate!
-
-Revision History:
-        1-1-2017: File Created                              [jbuhagiar]
-
---*/
+/** @file
+ *
+ *  Functions and structures to load a texture from an arc section
+ *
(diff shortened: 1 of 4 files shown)


Quaker762:
I'm going to commit this tomorrow because I need to go to bed. If you want to, you can do up all of the doxygen stuff, or leave it to me for tomorrow.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
-/*++
+/** @file
+ *
+ *  Functions and structures to load a texture from an arc section
+ *
+ *  \copyright 2016  Palm Studios
```
z33ky: We should choose either consistently `@tag` or `\tag`. I made the same error in my doxygen branch...

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
-        Defines data structures to load an manipulate a texture
+namespace sh3_graphics
+{
+    /**
+     * Full texture header (both batch and individual texture).
+     */
+    typedef struct
+    {
+        std::uint32_t batchHeaderMarker;          // This should be 0xFFFFFFFF to mark a header chunk
+        std::uint32_t unused1;                    // There are a lot of unused DWORDs, I assume to align everything nicely
```
z33ky: Perhaps it would be easier to understand byte-sized, instead of 4-byte sized, unused fields.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
-Abstract:
-        Defines data structures to load an manipulate a texture
+namespace sh3_graphics
+{
+    /**
+     * Full texture header (both batch and individual texture).
+     */
+    typedef struct
+    {
+        std::uint32_t batchHeaderMarker;          // This should be 0xFFFFFFFF to mark a header chunk
```
z33ky: Use `/**< This should be 0xFFFFFFFF to mark a header chunk */` to document this field in Doxygen.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
+#include <cstdint>
+#include <vector>
+#include <GL/gl.h>
 
-Abstract:
-        Defines data structures to load an manipulate a texture
+namespace sh3_graphics
+{
+    /**
+     * Full texture header (both batch and individual texture).
```
z33ky: @Add an empty comment line between `Full texture header` and the rest, because the first line is treated as a brief description by Doxygen.

Essentially, I think it should look like this:
```
   /** 
    *  Full texture header
    *  
    *  Both batch and individual texture.
    */
```

Also, I used two spaces after the asterisk, that aligns it better to the multi-line comment start IMO.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
+     *
+     * \note
+     *
+     * Defines a few bits of data and some functions to load in a texture
+     * from a SILENT HILL 3 .arc section.
+     */
+    class texture
+    {
+    public:
+    enum class result
```
z33ky: Deliberately not indented?

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
+    {
+        INVALID_CHUNK,
```
z33ky: My `enum class`es used CamelCase. We should decide on a consistent style.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
 
-Environment:
+    };
 
-Notes:
-        I'd like to thank Mike for all this, he put a lot of time into reverse engineering
-        all of the file types, including all of the textures, meshes and even the motion capture
-        skeletal animation that Konami captured. Thanks mate!
+        texture(sh3_arc& arcref){arc = arcref;}
+        texture(std::string& filename, sh3_arc& arcref){arc = arcref; Load(filename);}
```
z33ky: `filename` can be `const`.

source/SH3/graphics/texture.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * \note
+ *
+ * \return \c result indicating whether the read failed or not
+ *
+ */
+texture::result sh3_graphics::texture::Load(const std::string& filename)
+{
+    if(arc.LoadFile(filename, pixbuff) == ARC_FILE_NOT_FOUND)
+    {
+        die("sh3_graphics::texture::Load( ): Unable to find file %s!", filename.c_str());
```
z33ky, line 22: Should this not simply be encoded in `texture::result` as another error condition?

z33ky (changes requested):
To avoid storing a `sh3_arc` in `texture` you can simply pass it as another parameter to `Load` besides the filename.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
+        std::uint32_t unused6;
+        std::uint32_t unknown1;                   // Completely unknown, probably unimportant for now
+        std::uint32_t unknown2;                   // Usually 1!
+        std::uint32_t unused7[15];
+    } sh3_texture_header;
 
-Author:
-        Jesse Buhagiar
+    /**
+     * \class texture
```
z33ky: You don't need to manually specify this.

include/SH3/graphics/texture.hpp
```diff
@@ -1,56 +1,89 @@
...
+     *
+     * \brief Describes a logical texture and files to
```
z33ky: At least in the `Doxyfile` of my doxygen branch, I've enabled autobrief to take the first line as brief description, which would eliminate the need for the brief-tag.

source/SH3/graphics/texture.cpp
```diff
@@ -0,0 +1,36 @@
...
+ *  \date 2/1/2017
+ *
+ *  \author Jesse Buhagiar
+ */
+#include <SH3/graphics/texture.hpp>
+#include <SH3/system/log.hpp>
+
+using namespace sh3_graphics;
+
+/**
```
z33ky: Documentation should be in the header, because you're documenting the interface.

source/SH3/graphics/texture.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * \fn Load
+ *
+ * \brief Load a texture and return a result to it
+ *
+ * \note
+ *
+ * \return \c result indicating whether the read failed or not
+ *
+ */
+texture::result sh3_graphics::texture::Load(const std::string& filename)
```
z33ky: Missing documentation for filename parameter.

z33ky (changes requested):

include/SH3/graphics/texture.hpp
Quaker762: Okay, I usually use Upper case with Underscores for anything constant (defines and enums), but I'll leave it up to you to decide which is better.

include/SH3/graphics/texture.hpp
Quaker762: I don't quite understand what you mean?

include/SH3/graphics/texture.hpp
Quaker762: Okay, I like the `@tag` better, so let's stick to that.

include/SH3/graphics/texture.hpp
z33ky: Stuff in `class textures` are indented two levels (8 spaces), except for the `enum` definition, which is only indented one level (4 spaces).

include/SH3/graphics/texture.hpp
z33ky: There is no better, it's just about consistency. I've reserved UPPER_CASE for preprocessor macros in my C++ stuff and CamelCase for everything class-related, that is class, struct & enum names, member functions (`static`) and methods, member and class/struct (`static`) variables and enum values. Other stuff is camelCase, that is class/struct-less functions, parameters, global and local variables. This has the nice side-effect that you can easily have a parameter, say in the constructor, with the same name as a member variable, because it is capitalized differently. It is a rather uncommon convention though.

include/SH3/graphics/texture.hpp
z33ky: This is not 100% fixed. You're missing the `<`. Without it the documentation will be attached to the field following the comment, instead of the field preceding it.

include/SH3/graphics/texture.hpp
```diff
@@ -70,20 +72,28 @@ namespace sh3_graphics
...
 
-        result Load(const std::string& filename);
+    /**
+     * Load a texture and return a result to it
+     *
+     * @note
+     *
+     * @param filename - Virtual file path to file.
+     *
+     * @return \c result indicating whether the read failed or not
```
z33ky: Use `@ref` instead of `\c` so that Doxygen links the "result" to the type.

include/SH3/graphics/texture.hpp
```diff
@@ -70,20 +72,28 @@ namespace sh3_graphics
...
+     *
+     */
+        result Load(const std::string& filename, sh3_arc& arc);
```
z33ky: `arc` is not documented.

include/SH3/graphics/texture.hpp
```diff
@@ -70,20 +72,28 @@ namespace sh3_graphics
 
     };
 
-        texture(sh3_arc& arcref){arc = arcref;}
-        texture(std::string& filename, sh3_arc& arcref){arc = arcref; Load(filename);}
+        texture(){}
+        texture(std::string& filename, sh3_arc& arc){Load(filename, arc);}
```
z33ky: `filename` can be `const`.

z33ky (changes requested):
Just two additional things:

* `result` should be `load_result` to it's clear that the `result` belongs to the `Load` method.
* What's with the empty `@note`s?

Then just squash it again and I won't be bikeshedding any longer ;)

PR #39, opened 2017-01-03
Fixed assert problem

Quaker762:
sh3_arc_file::ReadData() was checking if the length of data to read was
GREATER than the max size of an int, hence causing an assertion fail.


z33ky (approved):

Quaker762:
@z33ky I'm getting another assertion fail in types.cpp, Line 180 after fixing the section loading (see #40 ). I don't really want to touch it atm in case I break anything else you've done.

z33ky:
Try #41.

PR #40, opened 2017-01-03
Fixed arc section loading

Quaker762:
Fixed a bug where type.cpp:LoadFile() was not looking in the /data/
folder or appending the ".arc" prefix to the string

### DIFF
--- a/source/SH3/arc/types.cpp
+++ b/source/SH3/arc/types.cpp
@@ -151,7 +151,8 @@ int sh3_arc::LoadFile(const std::string& filename, std::vector<std::uint8_t>& bu
 
     static_cast<void>(index); // avoid clang warning - index *is* initialized now
 
-    std::ifstream sectionFile(section->sectionName, std::ios::binary);
+    std::string sectionPath = "data/" + section->sectionName + ".arc";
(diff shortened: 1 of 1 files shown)


source/SH3/arc/types.cpp
```diff
@@ -151,7 +151,10 @@ int sh3_arc::LoadFile(const std::string& filename, std::vector<std::uint8_t>& bu
 
     static_cast<void>(index); // avoid clang warning - index *is* initialized now
 
-    std::ifstream sectionFile(section->sectionName, std::ios::binary);
+    std::string sectionPath = "data/" + section->sectionName;
+    sectionPath.insert(sectionPath.size() - 1, ".arc\0");
```
z33ky: `std::string sectionPath = "data/" + section->sectonName + ".arc";` should do.

z33ky (changes requested):

source/SH3/arc/types.cpp
Quaker762: The problem is there's a NUL terminator at the end of section->sectionName (part of the arc spec), which means the ".arc" part gets ignored

z33ky: I see. We should remove it when reading section->sectionName then.

PR #50, opened 2017-01-04
Work around gcc warning about [[noreturn]] main()

z33ky:


z33ky:
`enum class` is not implicitly castable to `int`, so we'd have to use `return static_cast<int>(exit_code::SUCCESS);`, which is a handful. I'm open to it though, especially because we don't need to write it often. We could maybe also define a short-hand, like `exit_cast`, though it doesn't save much.

Destructors not running could be bad. We could also switch to a plain `enum`, but we'd loose the scoping, which I'm quite fond of. We could work around it with a non-constructible struct or namespace with `constexpr` members. Anyways, discussion in #53.

PR #51, opened 2017-01-04
Remove arch.hpp and math.hpp

z33ky:


Quaker762:
I still think it's important to have a run-time check on Architecture/CPU model. Even though most people have SSE3/Core ix series processors, there could be a chance someone runs it on something really old (or maybe I'm just living in the 90's when you actually needed this stuff)

z33ky:
It depends on the build flags used which extensions the compiler uses though.
Like I said #20 the compiler can also emit x87 code. It can probably even to soft-float.

We could do something like
```
bool check_architecture()
{
    bool supported = true;
#ifdef __AVX2__
    supported &= __builtin_cpu_supports("avx2");
#endif
#ifdef __AVX__
    supported &= __builtin_cpu_supports("avx");
#endif
#ifdef __SSE4_2__
    supported &= __builtin_cpu_supports("sse4.2");
#endif
#ifdef __SSE4_1__
    supported &= __builtin_cpu_supports("sse4.1");
#endif
    ...
    return supported;
}
```
But I think if people's application crashes due to an illegal instruction, this should be straight forward to figure out. Ask for their processor type, check how the binary is built and you got the answer.

I guess what could be more interesting is changing implementations in runtime by compiling multiple versions with different CPU features and using a dispatch table selected depending on the available features.
Seems excessive, especially when people can theoretically recompile the code with whatever features they want. Practically not all people have the required knowledge to do so, but then you could provide different binaries for the prominent CPU features.

Quaker762:
Very true, I suppose the necessity of CPU checking is more suited towards closed source applications, where the chance of fixing a CPU related problem is practically 0 (unless there's a target for the architecture ala 90's AMD and Intel differences, **cough**3DNow!**cough**)

> I guess what could be more interesting is changing implementations in runtime by compiling multiple versions with different CPU features and using a dispatch table selected depending on the available features.

This is what I was originally trying to achieve in `math.hpp` with the inline assembly and `arch.hpp`. Check which is supported on function call and execute accordingly. However, I agree that is very excessive when people can target any processor.

Issue #43, opened 2017-01-03
sh3_arc::LoadFile( ) seeming reads nothing


Quaker762:
Printing out the buffer in it's entirety with 

`    for(int i = 0; i < 2048; i++)
     std::printf("0x%x ", buffer[i]);`

on Line 190-191 in types.cpp results in only 0s being printed.  This is the case when passing a vector reference to it in texture as well. 

Also, how would I copy values from the read `std::vector` into a struct? Google isn't really giving me any answers.

z33ky:
Could be an off-by-one wrt the index. I do get a 0xffffffff if I use `(index - 1) * sizeof(fileEntry)` instead of `index * sizeof(fileEntry)` when seeking into the file in `sh3_arc::LoadFile`.

z33ky:
> Also, how would I copy values from the read std::vector into a struct?

Completely missed that one. There should be an API like `sh3_arc_file` for this.

Quaker762:
Okay, gotcha. So the caller can't just copy a section of the vector into struc? (sorry, I'm so used to C file handling/buffers haha)

z33ky:
It's possible, just `std::copy(begin(vector), end(vector), reinterpret_cast<char*>(&struct));`, but why go through the vector when you can load the data directly to the struct?

Quaker762:
So we should write a few API calls (like the arc file stuff) that can read n bytes of data from the loaded file (which is in a vector you pass to LoadFile) into the struct, yeah? Excuse my ignorance haha.

z33ky:
I was thinking about dropping the `vector`. `sh3_arc_file::ReadString()` uses a `std::vector` internally because you cannot just copy bytes directly into a `std::string`. `std::string::data` does return `char*`, but it's undefined behaviour to modify that memory.
But if you look at `sh3_arc_file::ReadObject()`, you can see that eventually `gzread` will write directly to the struct's memory. I imagine the same thing for `LoadFile()`, where you will `std::ifstream::read` directly into the struct.

Quaker762:
Alright, sounds good. It'll probably make life a lot easier being able to read into multiple types.

Quaker762:
After doing some more testing, it seems this has been fixed. I am getting correct output from the read vector. 

I've done a bit of research, and I think the vector may be a good idea to keep. Copying from a vector will definitely come in handy when allocating buffers, because we won't have to manually probe/write a function to query the file size (though this might be a good idea to have.)

[...]

PR #55, opened 2017-01-05
Initial implementaiton of Virtual Files

Quaker762:
Title explains it all. These are supposed to somewhat mimic C file IO,
as some file types will need to be seeked (Mesh files, animation files
etc). I think this makes a nice way to handle Virtual Files from arc
sections.

However, I feel the std::vector buffer may be a waste of space, and
there could be a MUCH better way to do this without buffering the entire
file (a HUGE waste of memory).

@z33ky There's probably a much better C++ whizbang way to do this (without the chance of any segfaults + more abstractions), but I think this is pretty sound for now, at least for the file types that I know about, especially when it comes to buffering sound.

### DIFF
--- a/include/SH3/arc/vfile.hpp
+++ b/include/SH3/arc/vfile.hpp
@@ -1,0 +1,98 @@
+/** @file
+ *
+ *  Class and data structures describing a Virtual File inside a SILENT HILL 3 arc section.
+ *
+ *  @copyright 2016  Palm Studios
+ *
+ *  @date 5-1-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#ifndef VFILE_HPP_INCLUDED
+#define VFILE_HPP_INCLUDED
+
+#include "SH3/arc/types.hpp"
+#include "SH3/error.hpp"
+#include <ios>
+
+/**
+ *  Opens a handle to a Virtual File contained within arc.arc
+ *
+ *  After the handle to the file is open, it is possible to seek and rewind
+ *  to a specific part of the file using @ref Seek().
+ *
+ *  Each read sets the @ref read_error indicating whether a partial read was performed,
+ *  or that the end of file was encountered (meaning that @ref fpos `+ len` was equal to @ref fsize.
+ *
+ */
+struct sh3_arc_vfile final
+{
+public:
+    /**
+     *  Failure codes for @ref ReadData.
+     */
+    enum class load_result
+    {
+        SUCCESS,
+        END_OF_FILE,
+        PARTIAL_READ,
+    };
+
+    struct read_error final : public error<load_result>
+    {
+    public:
+        std::string message() const;
+    };
+
+
+    sh3_arc_vfile(sh3_arc& mft, const std::string& fname): fpos(0)
+    {Open(mft, fname);}
+
+    /**
+     *  Read @c len bytes of data into a destination buffer.
+     *
+     *  @param destination      Pointer to write read values into.
+     *  @param len              Number of bytes to read from the file.
+     *  @param e                @ref read_error from this operation.
+     *
+     *  @returns Number of bytes read (which should be len)
+     */
+    std::size_t ReadData(void* destination, std::size_t len, read_error& e);
+
+    /**
+     *  Rewind this file to the beginning (set fpos to 0).
+     */
+    void Rewind(){fpos = 0;};
+
+    /**
+     *  Seek to a certain position in the file.
+     *
+     *  @param pos      Position to seek to.
+     *  @param origin   The origin we want to seek from.
+     */
+     void Seek(long pos, std::ios_base::seekdir origin);
+
+
+private:
+    std::uint32_t   fpos;           /**< Current file position */
+    std::size_t     fsize;          /**< Size of this file inside the arc section in bytes */
+    std::string     fname;          /**< The name of this file (taken from arc.arc) */
+    bool            open = false;   /**< Is this file handle currently open? */
+
+    std::vector<std::uint8_t> buffer; /**< Buffer that @ref ReadData() reads from */
+
+    /**
+     *  Open a handle to a virtual file.
+     *
+     *  @param mft      The @ref sh3_arc Master File Table, arc.arc.
+     *  @param fname    The name of the file we want to open.
+     *
+     *  @note If the file is already open, this function returns false.
+     *
+     *  @returns @c true if the file was found, @c false if not.
+     */
+    bool Open(sh3_arc& mft, const std::string& fname);
+};
+
+
+#endif // VFILE_HPP_INCLUDED
--- a/source/SH3/arc/vfile.cpp
+++ b/source/SH3/arc/vfile.cpp
@@ -1,0 +1,92 @@
+/** @file
+ *
+ *  Implementation of sh3_arc_vfile.hpp
+ *
+ *  @copyright 2016  Palm Studios
+ *
+ *  @date 5-1-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include <SH3/arc/vfile.hpp>
+#include <SH3/arc/types.hpp>
+#include <SH3/system/log.hpp>
+#include <cstring>
+#include <cassert>
+
+bool sh3_arc_vfile::Open(sh3_arc& mft, const std::string& fname)
+{
+    if(open) return false;
+
+    /*
+        Load the file from the section it is, and set our local fsize to
+        it (so we know how large it is without probing) though most headers contain the size of the
+        full file
+    */
+    if((fsize = mft.LoadFile(fname, buffer)) == ARC_FILE_NOT_FOUND)
+    {
+        open = false;
+        return open;
+    }
+
+    open = true;
+    return open;
+}
+
+std::string sh3_arc_vfile::read_error::message() const
+{
+    std::string error;
+    switch(result)
+    {
+    case load_result::SUCCESS:
+        error = "Success";
+        break;
+    case load_result::END_OF_FILE:
+        error = "EOF";
+        break;
+    case load_result::PARTIAL_READ:
+        error = "Partial Read";
+        break;
+    }
+    return error;
+};
+
+void sh3_arc_vfile::Seek(long pos, std::ios_base::seekdir origin)
+{
+    if(origin == std::ios_base::beg)
+    {
+        fpos = pos;
+    }
+    else if(std::ios_base::cur)
+    {
+        fpos += pos;
+    }
+
+    // Log if there is an attempt at something naughty
+    if(pos > fsize)
+    {
+        Log(LogLevel::WARN, "sh3_arc_vfile::Seek( ): Attempt to seek to position > fsize! Setting fpos to 0!");
+        fpos = 0;
+    }
+}
+
+std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_error& e)
+{
+    if(len >= buffer.size())
+    {
+        e.set_error(load_result::END_OF_FILE);
+    }
+    else if(len < fsize)
+    {
+        e.set_error(load_result::PARTIAL_READ);
+    }
+
+    assert(fpos > fsize); // This should never EVER happen. Terminate if it does.
+
+    std::size_t nbytes = std::min(len, fsize - fpos);
+    std::memcpy(destination, buffer.data() + fpos, nbytes);
+
+    fpos += nbytes; // Increment the position we are at in this file
+
+    return nbytes;
+}


Quaker762:
Oh, and I should note I haven't implemented a few things yet (on purpose), but if you're okay with this type of file handing, I'll finish them up. Also there's probably a few places where the buffer can be read past it's length, which I'll need to fix up.

include/SH3/arc/vfile.hpp
```diff
@@ -0,0 +1,100 @@
...
+        void set_error(sh3_arc_vfile::read_result res){result = res;};
+
+        std::string message() const;
+
+    private:
+        read_result result = read_result::SUCCESS;
+    };
+
+
+    sh3_arc_vfile(sh3_arc& mft, std::string fname){Open(mft, fname);fpos = 0;}
```
z33ky: `fname` can be `const std::string&`.

`fpos` can be initialized through member initialization ala `open`.

include/SH3/arc/vfile.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     *  Open a handle to a virtual file.
+     *
+     *  @param mft      The @ref sh3_arc Master File Table, arc.arc.
+     *  @param fname    The name of the file we want to open.
+     *
+     *  @returns @c true if the file was found, @c false if not.
+     */
+    bool Open(sh3_arc& mft, std::string fname);
```
z33ky: `fname` can be `const std::string&`.

Documentation should specify what happens when you call `Open()` on an already opened file.

include/SH3/arc/vfile.hpp
```diff
@@ -0,0 +1,100 @@
...
+     *  Closes the handle to this file, and empties the buffer vector
+     */
+    void Close(){fpos = 0; open = false; buffer.clear();};
+
+    /**
+     *  Read @c len bytes of data into a destination buffer.
+     *
+     *  @param destination      Pointer to write read values into.
+     *  @param len              Number of bytes to read from the file.
+     *  @param e                Read error class.
```
z33ky: This should `@ref read_error`.

include/SH3/arc/vfile.hpp
```diff
@@ -0,0 +1,100 @@
...
+     */
+    void Rewind(){fpos = 0;};
+
+    /**
+     *  Seek to a certain position in the file.
+     *
+     *  @param pos      Position to seek to.
+     *  @param origin   The origin we want to seek from.
+     */
+     void Seek(long pos, int origin);
```
z33ky: `origin` should be `std::ios_base::seek_dir`.

z33ky:
To avoid caching the vfile contents in an std::vector, perhaps it should just store a reference to the MFT file then.
For convenience, we could also mmap the arc files. Boost.IOStreams has a cross-platform `mapped_file` [class](http://www.boost.org/doc/libs/release/libs/iostreams/doc/classes/mapped_file.html#mapped_file).

Alternatively, with the `resource` interface I just proposed over at #43, we wouldn't need `sh3_vfile` and could just let `sh3_arc` read into a `std::vector` and just keep the vector instead of having to copy bytes out again.

Quaker762:
Would it still be possible to seek the file without caching it though? Also, isn't the use of Boost considered bad, as it's so bloated? Or am I thinking of incorrect use of the library?

Also, see my comment in #43

include/SH3/arc/vfile.hpp
Quaker762: I really need to remember to have my strings as const references...

z33ky:
> Would it still be possible to seek the file without caching it though?

Sure. Memory mapping means you get access to the file through a `char*`. Seeking is then just pointer arithmetic.

> Also, isn't the use of Boost considered bad, as it's so bloated?

It's over-engineered for many use cases. It can be a pain, it can be a saint. Some of Boost's libraries will increase the compile-time more so than simpler versions doing the same thing. I would suggest just trying it out and considering to replace stuff that slows the compilation down in an unacceptable manner. We can also try precompiled headers.
As far as the binary goes, you only pay for what you use. Boost gets well tested an a range of platforms and compilers and the libraries are peer-reviewed and improved upon incrementally.
Apart from Boost.IOStreams for this I'd also look into [Boost.Range](http://www.boost.org/doc/libs/release/libs/range/) (just generally convenient) and [Boost.Filesystem](http://www.boost.org/doc/libs/release/libs/filesystem).

include/SH3/arc/vfile.hpp
```diff
@@ -67,7 +65,7 @@ struct sh3_arc_vfile final
      *
      *  @param destination      Pointer to write read values into.
      *  @param len              Number of bytes to read from the file.
-     *  @param e                Read error class.
+     *  @param e                @ref read_error that this file has.
```
z33ky: "`read_error` from this operation" would be more accurate.

include/SH3/arc/vfile.hpp
```diff
@@ -53,9 +49,11 @@ struct sh3_arc_vfile final
      *  @param mft      The @ref sh3_arc Master File Table, arc.arc.
      *  @param fname    The name of the file we want to open.
      *
+     *  @note If the file is already open, this function does nothing but returns @ref open.
```
z33ky: Shouldn't it perhaps return `false` because a different file could be open?
Actually, this method maybe should be private, since the Constructor always already opens the vfile. That way it should never happen that `Open()` is called twice.

source/SH3/arc/vfile.cpp
```diff
@@ -31,5 +57,16 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
-    std::memcpy(destination, buffer.data() + fpos, len);
+    // Prevent memcpy from reading past the vector
+    if(fpos + len > fsize)
+    {
+        std::size_t nbytes = len - fpos
+        std::memcpy(destination, buffer.data() + fpos, nbytes);
+    }
+    else
+    {
+        std::memcpy(destination, buffer.data() + fpos, len);
```
z33ky: This almost duplicated line isn't needed if `std::memcpy` is always called with `nbytes`, which is set to the appropriate size.

source/SH3/arc/vfile.cpp
```diff
@@ -31,5 +57,16 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
         e.set_error(read_result::PARTIAL_READ);
     }
 
-    std::memcpy(destination, buffer.data() + fpos, len);
+    // Prevent memcpy from reading past the vector
+    if(fpos + len > fsize)
```
z33ky: Unlikely, but take care of overflow on fpos + len.

include/SH3/arc/vfile.hpp
Quaker762: Very true, I was considering making this private, but I thought it might be better to give more control. Making it private makes it much safer though.

source/SH3/arc/vfile.cpp
Quaker762: Could you elaborate on this, I don't quite understand what you mean?

z33ky: `fpos + len` may overflow the datatype. You could do `len = std::min(len, fsize - fpos);` instead though; fpos should never be larger than fsize (you could `assert()` that), otherwise we get an underflow.

source/SH3/arc/vfile.cpp
```diff
@@ -31,7 +33,7 @@ bool sh3_arc_vfile::Open(sh3_arc& mft, const std::string& fname)
         return open;
     }
 
-    return open;
+    return false;
```
z33ky: You can avoid indentation of the previous code with an early return (`if(open) return false;`).

Quaker762: Completely missed that, 2am code lmao.

Quaker762:
@z33ky  Is this ready for a merge yet?

z33ky:
I think the implementation for `read_error::message()` is missing (or I missed it?) and documentation for the `sh3_arc_vfile` class would be good I think.
Apart from that it looks good. I should probably write a generalized version of the `read_error` class, which takes an `enum` via template parameter, to reduce boilerplate code.

Quaker762:
Oops, it seems I am missing the implementation of `read_error::message()`... I'll get onto that and the documentation right away. 
Sounds good, that would save us having to add it to every class that implements some form of IO.

Quaker762:
@z33ky This should be all ready to go.

Sorry about the lack of commits lately by the way, I've had work for the past 3 days, as well as accidentally formatting my Media drive......

include/SH3/arc/vfile.hpp
```diff
@@ -14,6 +14,24 @@
...
 #include <ios>
 
+/**
+ *  Opens a handle to a 'virtual file' within an arc section.
+ *
+ *  After the handle to the file is open, it is possible to seek and rewind
+ *  to specific parts of the file using @ref Seek().
+ *
+ *  Each read sets the @ref read_error to indicate whether a partial read was performed
+ *  or that the end of file was encountered (meaning that @ref fpos + len was equal to @ref fsize.
```
z33ky: ``@ref fpos `+ len` `` puts the "+ len" parts in typewriter font, which is normally used for code snippets.
Sadly there doesn't seem to be a way to use `@ref` within that, so keep it outside the quotation.

include/SH3/arc/vfile.hpp
```diff
@@ -14,6 +14,24 @@
 #include "SH3/arc/types.hpp"
 #include <ios>
 
+/**
+ *  Opens a handle to a 'virtual file' within an arc section.
+ *
+ *  After the handle to the file is open, it is possible to seek and rewind
+ *  to specific parts of the file using @ref Seek().
+ *
+ *  Each read sets the @ref read_error to indicate whether a partial read was performed
```
z33ky: "Each read returns a @ref read_error indicating..." is more accurate.

include/SH3/arc/vfile.hpp
```diff
@@ -14,6 +14,24 @@
...
+ *  or that the end of file was encountered (meaning that @ref fpos + len was equal to @ref fsize.
+ *
+ *  Each time a file is opened, the @ref sh3_arc MFT must be passed to the constructor so that there is access to
+ *  the file table.
+ *
+ *  @note The ENTIRE file contents is stored in an @c std::vector. You should make CERTAIN to close any files
```
z33ky: This can be removed, because the destructor will do this.

z33ky (changes requested):
The `read_error` can use the `error` template that was just merged.
And we might not need `Close`, for the reasons given for the `@note` in the class documentation.

z33ky:
> Sorry about the lack of commits lately by the way, I've had work for the past 3 days, as well as accidentally formatting my Media drive......

No sweat. I wanted to work on the `resource` thingy mentioned in one of the issues and I have less good reasons for not having done so, which might involve alcohol.

And squash the commits. Most of them are not useful individually.

include/SH3/arc/vfile.hpp
Quaker762: I'll keep `Close()` for now, but yeah it'll probably get the axe.

Quaker762:
Okay I've seriously fucked my branch up big time. Can you possibly recommend a fix (I'll refork when this gets merged)

Quaker762:
> ... I have less good reasons for not having done so, which might involve alcohol.

100% relatablehaha

z33ky:
I squashed the commits on my branch. The commit message needs to be cleaned up (use `git commit --amend`), that that's it.
I've removed the `Close()` method while I'm at it, as I suggested in a comment.
Edit: Sorry, didn't see your response before doing that. You can re-add it if you want. Just change the file and after `git add`ing it, do `git commit --amend --no-edit`.

You can get my stuff by checking out your vfile branch and doing `git remote add z33ky git@github.com:z33ky/sh3redux.git && git fetch z33ky vfile && git reset --hard FETCH_HEAD`. If you have local changes, do a `git stash` before and `git stash pop` afterwards.

Quaker762:
It seems I don't have permission to fetch from your fork.

z33ky:
Change `git remote add z33ky git@github.com:z33ky/sh3redux.git` to `git remote set-url z33ky https://github.com/z33ky/sh3redux`.

Quaker762:
Done

z33ky (approved):

z33ky:
You might still want to clean up the commit message (`git commit --amend`), currently it's just the commit messages of the previously 3 (4?) commits in one text.

Quaker762:
Looks good to go now (finally, hahaha)

PR #56, opened 2017-01-06
Add assert.hpp

z33ky:
~Do not merge this yet. I will add documentation.~
I'm happy with the implementation though, so this can already be reviewed.
To build you will need to `git submodule update --init` to get the debugbreak-submodule.


z33ky:
As with standard `assert()`, side-effects in the assertion are not present in release-mode. I have thought about adding a `ASSERT_EXEC_ALWAYS()` macro for this, which will not assert in release-mode, but will execute the code producing the conditional, so that side-effects are always evaluated.
Alternatively we could go the safe route and make this behaviour the default, perhaps a `ASSERT_NOEXEC_RELEASE()` for expensive checks that we want to avoid in release-mode.

Additionally I thought about adding a `ASSERT_ALWAYS()` that will also assert in release-mode, and a `ASSERT_RELEASE()` that can be used to replace expensive checks in debug-mode with less expensive ones in release-mode.

z33ky:
Some fixes, minor changes and documentation added.

Quaker762:
Looks great, this'll speed up fixing errors easily, especially when we can get builds out to people (a long way away...)

PR #61, opened 2017-01-12
Fixed assert problem in 'vfile.cpp'

Quaker762:
Totally forgot that the assert error is thrown if 0 is returned from the
expression.... Ooops!

### DIFF
--- a/source/SH3/arc/vfile.cpp
+++ b/source/SH3/arc/vfile.cpp
@@ -76,14 +76,14 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
     {
         e.set_error(load_result::END_OF_FILE);
     }
-    else if(len < fsize)
+
+    assert(fpos <= fsize); // This should never EVER happen. Terminate if it does.
+
+    if((std::size_t nbytes = std::min(len, fsize - fpos)) != len)
     {
         e.set_error(load_result::PARTIAL_READ);
     }
 
-    assert(fpos > fsize); // This should never EVER happen. Terminate if it does.
-
-    std::size_t nbytes = std::min(len, fsize - fpos);
     std::memcpy(destination, buffer.data() + fpos, nbytes);
 
     fpos += nbytes; // Increment the position we are at in this file


z33ky:
`fpos <= fsize` is safe, too. It'll just read nothing, but it's still sane behavior.

Quaker762:
Okay that should do it. I've taken into account `fpos+len`, because otherwise it'll still read past the buffer.

z33ky:
Actually, this is what the `std::min()` call is for. `nbytes` is `len` capped by the amount of remaining bytes.
Edit: we should set the `read_error` to `PARTIAL_READ` if is is being capped though.

Quaker762:
I completely forgot that was in there... I'll fix that ASAP.

source/SH3/arc/vfile.cpp
```diff
@@ -81,9 +81,13 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
         e.set_error(load_result::PARTIAL_READ);
     }
 
-    assert((fpos + len) <= fsize); // This should never EVER happen. Terminate if it does.
+    assert(fpos <= fsize); // This should never EVER happen. Terminate if it does.
+
+    if((std::size_t nbytes = std::min(len, fsize - fpos)) == len)
```
z33ky: That should be `!= len`. Also I think this makes lines 79-82 redundant, as that case should be covered by this check.

source/SH3/arc/vfile.cpp
```diff
@@ -81,9 +81,13 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
+	{
```
z33ky: You switched to tabs instead of spaces here.

z33ky (changes requested):

z33ky:
Also, don't forget so squash :)
The fix of the assert and the adaption of the error can be done in separate commits, if you want to try your hand on more interactive rebasing. See the "edit" command.

source/SH3/arc/vfile.cpp
Quaker762: Whoops, I opened the file up in a different editor to write this line haha

z33ky (approved):

PR #63, opened 2017-01-13
Fixes for vfile.cpp

z33ky:


source/SH3/arc/vfile.cpp
```diff
@@ -55,15 +60,22 @@ void sh3_arc_vfile::Seek(long pos, std::ios_base::seekdir origin)
 {
     if(origin == std::ios_base::beg)
     {
-        fpos = pos;
+        fpos = static_cast<decltype(fpos)>(pos);
     }
-    else if(std::ios_base::cur)
+    else if(origin == std::ios_base::cur)
     {
-        fpos += pos;
```
z33ky, line 62: This line might be fine with C++'s convertion rules and unsigned integer overflow. Need to investigate...

z33ky: See the following amended change...

Quaker762: Alright, do you want me to merge or look this up first?

source/SH3/arc/vfile.cpp
```diff
@@ -23,13 +23,18 @@ bool sh3_arc_vfile::Open(sh3_arc& mft, const std::string& fname)
         it (so we know how large it is without probing) though most headers contain the size of the
         full file
     */
-    if((fsize = mft.LoadFile(fname, buffer)) == ARC_FILE_NOT_FOUND)
+    int size = mft.LoadFile(filename, buffer);
```
Quaker762, line 26: Haha, can you tell where I learned C (*cough*K&R*cough*)

Quaker762 (approved):

source/SH3/arc/vfile.cpp
z33ky: It's a perfectly fine technique, however clang complains about some implicit casts, which I avoid by buffering the value in a local `int`.

source/SH3/arc/vfile.cpp
z33ky: The comment below gives a rough explanation why it's fine. Feel free to verify that it's true though.

source/SH3/arc/vfile.cpp
Quaker762: Ah, I see. g++ doesn't seem to complain about this. Maybe I need to enable more warnings in Code::Blocks

PR #64, opened 2017-01-15
Updated codeblocks project

Quaker762:
Updated the codeblocks project (it was HORRIBLY out of date and didn't
have correct compiler warnings enabled).


SILENT HILL 3 Redux.cbp
```diff
@@ -19,6 +19,9 @@
 					<Add option="-masm=intel" />
```
z33ky: You should remove this flag as I've done in e54ac6aa919b418edcb40d1eeb62713fe1562e55 for the `CMakeLists.txt`.

Quaker762: Hmm, it mustn't have saved my project file properly, as I removed that flag. I'll fix it now.

PR #65, opened 2017-01-16
Initial declaration of the message class

Quaker762:
Initial data structure and function declarations created. This class
represents a message in SILENT HILL 3. See the documentation in 'msg.hpp' for more.

There's more to come for this, I'll push some more commits as they come.

This format is a little tricky, I'm not quite sure how the game determines which message to display, though Mike seemed to think they were in the `*.kg1,` `*.kg2` files. If not, we'll have to ID doors etc and hard code IDs ourselves (hopefully not). Time for some reverse engineering!


SILENT HILL 3 Redux.cbp
```diff
@@ -85,31 +87,47 @@
...
+		<Unit filename="include/SH3/arc/vfile.hpp" />
+		<Unit filename="include/SH3/error.hpp" />
 		<Unit filename="include/SH3/graphics/model.hpp" />
 		<Unit filename="include/SH3/graphics/texture.hpp" />
-		<Unit filename="include/SH3/math.hpp" />
 		<Unit filename="include/SH3/system/config.hpp" />
 		<Unit filename="include/SH3/system/exit_code.hpp" />
 		<Unit filename="include/SH3/system/glcontext.hpp" />
 		<Unit filename="include/SH3/system/log.hpp" />
+		<Unit filename="include/SH3/system/msg/msg.hpp" />
```
z33ky: Please also add this to `source/CMakeLists.txt`.

include/SH3/system/msg/msg.hpp
```diff
@@ -0,0 +1,73 @@
...
+ *  @date 12-1-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#ifndef MSG_HPP_INCLUDED
+#define MSG_HPP_INCLUDED
+
+#include <cstdint>
+
+#define SH3_2_ASCII(x) ((x) - 32) /**< Turns a SILENT HILL 3 msg character into an ASCII character */
```
z33ky: What isn't part of the public interface should go to the cpp file.

I'd also prefer this to be a function.

include/SH3/system/msg/msg.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#define SH3_MSG_BARS        0xFA /**< Tells the font manager to draw 3 lines of black bars across the screen (???) */
```
z33ky: These look like they should be an enum.

include/SH3/system/msg/msg.hpp
```diff
@@ -0,0 +1,73 @@
...
+#define SH3_MSG_BREATHE     0xFB /**< Tells the font manager to breathe the text in and out from white to red and back again */
+#define SH3_MSG_NEWLINE     0xFD /**< Tells the font manager to draw a newline for the next part of this message */
+#define SH3_ATTR_START      0xFF /**< Signals that the next byte will be a text attribute */
+
+#define SH3_MSG_WHITECLR    0x00 /**< Tells the font manager to draw this message in white or transparent (?) */
```
z33ky: Same (enum).

source/SH3/system/msg/msg.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+using namespace sh3_system;
+
+void sh3_msg::Load(sh3_arc_vfile& file)
+{
+
+    std::uint8_t c;
+    std::uint8_t c2;
+    sh3_arc_vfile::read_error e;
+
```
z33ky: We should `msgString.reserve()` an approximate message length, by adding an accessor for `sh3_arc_vfile::fsize`.
`msgString.shrink_to_fit()` can be used to free any excess memory.

source/SH3/system/msg/msg.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+#include "SH3/SH3/system/msg/msg.hpp"
+
+using namespace sh3_system;
+
+void sh3_msg::Load(sh3_arc_vfile& file)
+{
+
+    std::uint8_t c;
+    std::uint8_t c2;
```
z33ky: An array or struct that can be read in at once would be cleaner.

source/SH3/system/msg/msg.cpp
```diff
@@ -0,0 +1,40 @@
...
+    sh3_arc_vfile::read_error e;
+
+    // While we are reading a character
+    while(file.ReadData(&c, 1, e))
+    {
+        file.ReadData(&c2, 1, e); // Read in a second character (attribute/terminator check)
+
+        if(c == SH3_ATTR_START && c2 == SH3_ATTR_START) // If BOTH characters == 0xFF, this message has been fully copied.
```
z33ky: This check can be moved into the `true`-branch of the next one.

Quaker762:
Alright I'll get to work on the changes for this and #67 asap (i.e tomorrow)

include/SH3/system/msg/msg.hpp
Quaker762: What would the difference be between it being a function as compared to a macro?? It seems like a bit of a waste imho.

source/SH3/system/msg/msg.cpp
Quaker762: > We should msgString.reserve() an approximate message length, by adding an accessor for sh3_arc_vfile::fsize.

You mean something like getFsize() yeah? The only problem I see is all the messages are stashed in the one file, so the size would be pretty large.

include/SH3/system/msg/msg.hpp
Quaker762: Is there any particular benefit besides having the enum::type usage?

If I do shove them in an enum, I only have to define the first value and the compiler will work out the others in order, yes?

include/SH3/system/msg/msg.hpp
z33ky: I could say type safety, but it'd be difficult to unintentionally misuse a macro that has "ASCII" in its name.
There's not much to debug here, so we don't really need this to appear in a debugger.
I also wouldn't expect much code to be added here...
So there is little to no technical reason why a function would be *better*, just a stylistic choice really. And there is no waste (?) except in debug mode, as it is certainly inline with optimizations enabled.

include/SH3/system/msg/msg.hpp
z33ky: Scoping, which you mentioned, but also type safety are the reasons I prefer enums over either `#define` or `constexpr` constants.
It you `switch` against enum values, compilers can also check that you cover all possible values (barring casting an invalid value to the enum type of course).

source/SH3/system/msg/msg.cpp
z33ky: Ah, makes sense. We could reserve some amount representative of one of the longer, average message sizes to avoid some memory allocation overhead. Or since we buffer the whole file anyway we could trivially expose a function that looks for some pattern and returns the offset, but since this might change I'd perhaps just leave it as is and plunk down a `//TODO: can we reserve() some space?`.

Quaker762: Sure thing, we'll have to revisit this one, hopefully when we can actually get something on the screen. 

I'm not entirely certain how the original game actually tells what message to load, as all the messages are stashed in the file... I (and Mike) think that one of the unknown file types contains some kind of interaction data, which tells which ID to render.

include/SH3/system/msg/msg.hpp
Quaker762: Okay sure thing. The waste I was referring to was a few clock cycles, but I suppose they'd both get compiled into the same thing. 

It's probably a good idea to have it as a function so we at least have the debugging symbol in case something does screw up (and I'm willing to bet something will [laughs]).

include/SH3/system/msg/msg.hpp
Quaker762: Alright, I've moved all of them to an enum.

z33ky:
I think a `std::bitset` with an `enum` for indexing would be a good fit for the attributes.

Quaker762:
@z33ky I just looked this up and it seems perfect for what we need. The only thing I'm still thinking of how we get working is when the attribute switches (seen here https://i.imgur.com/b2mKkaLh.jpg). Should we draw character by character into a buffer and then display? (Which I imagine is slow)

z33ky:
I imagine we would do text rendering via a shader. For this case it would be good to spend some memory to store attributes *per character*, so that drawing can be stateless.
For this we could just define our own `attributed_character` class, storing the `char` and a `std::bitset` and use `std::basic_string<attributed_character>`.
But `std::bitset` does not make guarantees about its internal layout I believe, which means that we would have to serialize it to upload it to the graphics hardware. We should then just use `std::uint8_t` and some bit-arithmetic instead - then we can just pass it to `glBufferData` - or whatever - as is. And wrap it in a `std::bitset`-like interface if we're feeling fancy.

Quaker762:
> spend some memory to store attributes per character

That's actually a good idea, so on `Load()` when the current attribute switches we store that in the `attributed_character` struct? 

> But std::bitset does not make guarantees about its internal layout I believe,

I see, so that'll probably best be the best option then? That or we can just send the entire struct to the GPU and decode the attribute in the shader?

z33ky:
> That's actually a good idea, so on `Load()` when the current attribute switches we store that in the `attributed_character` struct?

Not just when it switches. In your example screenshot, every character up to the "shopping mall map" has no attribute, then every character has the green color-attribute with the exception of the final ".", which again has no attribute.
This way the shader only needs the index into the buffer to find out both the character to draw and its attribute.
That said, even figuring out the index to the buffer from a pixels x and y coordinates is not fully straight forward. We should look at examples of rich text rendering done in GLSL.

> I see, so that'll probably best be the best option then? That or we can just send the entire struct to the GPU and decode the attribute in the shader?

This is precisely why we should refrain from `std::bitset`, because it probably cannot be just sent to the GPU. Once there, a GLSL shader needs to figure out how to read the block of data we sent, and while we know the first part is a `char`, we don't know how to treat the bytes making up the `std::bitset`.

z33ky:
It seems I am mistaken with my expectation on how to draw text with OpenGL.
I thought you'd upload a complete string and essentially render it with a single `glDraw*()` call.
However it seems common to just set up vertices and shader variables and use a `glDraw*()` call *for each character*, not for the whole string at once.
In this case, we can store the string using a regular `std::string` and perhaps encode attributes via a `std::vector` of `struct{ std::bitset<N> attributes; std::size_t length; };`, where the `attributes` are active for `length` characters.
Embedding the attributes in the string, as is done in the data files, isn't the worst option either though.

Quaker762:
> In your example screenshot, every character up to the "shopping mall map" has no attribute, then every character has the green color-attribute with the exception of the final ".", which again has no attribute.

As ridiculous as it sounds, [the attribute switches back to white right before the end of the message](http://i.imgur.com/co8vfeU.png)... This is a very VERY strange data format team SILENT came up with...

> That said, even figuring out the index to the buffer from a pixels x and y coordinates is not fully straight forward. We should look at examples of rich text rendering done in GLSL.

Couldn't we just use [FTGL](http://ftgl.sourceforge.net/docs/html/ftgl-tutorial.html) with texture fonts, and then just create a new string object every time we need to (this way we don't have to send data to any shader). The only drawback is per character attributing, which I don't think it supports unless we render character by character, checking the `attributed_character`'s attribute every call? Considering we only need to draw to the buffer once at any given time (scene load, interaction, cutscene [which I still have absolutely 0 idea where the data is stored for..]), I don't see too much of a waste of CPU time. 

> However it seems common to just set up vertices and shader variables and use a glDraw*() call for each character, not for the whole string at once.

This could definitely work too, though I've never actually seen any of this done with Shaders, I got pointed to FTGL (which is probably doing this exact thing, though on the fixed pipeline).

> In this case, we can store the string using a regular std::string and perhaps encode attributes via a std::vector of struct{ std::bitset<N> attributes; std::size_t length; };, where the attributes are active for length characters.
Embedding the attributes in the string, as is done in the data files, isn't the worst option either though.

Either or; it's probably easier to store it like you've said, but then again we just 'stream & switch' when we hit them during runtime (or load time?).

EDIT: As a side note, I still have no idea where in the hell the English font is actually stored, and I know it's there somewhere, considering it's so aliased and blurry. If we are to use FTGL with a `.ttf` font, the closest thing to it is, believe it or not, [Times New Roman](https://www.reddit.com/r/silenthill/comments/3dgj8u/does_anybody_know_how_to_gain_access_to_what_i/?st=iybkyyzd&sh=56665ce5), except modified. Now that I think about it, it could be embedded in the .exe.

z33ky:
The OpenGL fixed-function pipeline was removed in 3.something.
Rendering text using a `glDraw*()` per character is quite simple really. We just upload a texture atlas that contains every character and to render text, you loop through the string you want to display, set the coordinates to the character in the atlas for the shader, which just blends the pixels from the texture.

Quaker762:
Okay that sounds simple enough. It's probably more efficient than using FTGL tbh.

We should get state management etc done asap so we can test some of this stuff out.

Quaker762:
Deprecated.

PR #67, opened 2017-01-17
Dumping files

Quaker762:
Added a function to dump virtual files to disk, so as to aid in reverse
engineering any file types that are currently unknown.


source/SH3/arc/vfile.cpp
```diff
@@ -100,3 +102,23 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
+    {
+        Log(LogLevel::WARN, "sh3_arc_vfile::Dump2Disk( ): Warning! Attempting to flush unopen or empty buffer to disk!");
+        return;
+    }
+
+    // Firstly, we need to convert the file path to just the filename.
+    std::stringstream ss(fname);
+    std::string item;
+
+    while(getline(ss, item, '/')); // Throw out all strings except the last one (our file name)
```
z33ky: It'd be cheaper to look for the last `/` in the string (see `std::string::rfind`) and passing an iterator pair to `std::ofstream`'s constructor (`begin(fname) + pos` and `end(fname)`).

source/SH3/arc/vfile.cpp
```diff
@@ -100,3 +102,23 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
+
+    std::ofstream out_file(item);
+    for(int i = 0; i < buffer.size(); i++)
+        out_file << buffer[i];
```
z33ky: The "streaming operators" (`<<`, `>>`) are meant for formatted data. We're dealing with binary data, so use `std::ofstream::write` instead. And you don't need to do it item by item then, you can just `write(buffer.data(), buffer.size())`.

source/SH3/arc/vfile.cpp
```diff
@@ -100,3 +102,23 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
+        return;
+    }
+
+    // Firstly, we need to convert the file path to just the filename.
+    std::stringstream ss(fname);
+    std::string item;
+
+    while(getline(ss, item, '/')); // Throw out all strings except the last one (our file name)
+
+    std::ofstream out_file(item);
```
z33ky: Check if opening fails.

source/SH3/arc/vfile.cpp
```diff
@@ -100,3 +102,23 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
 
     return nbytes;
 }
+
+void sh3_arc_vfile::Dump2Disk()
+{
+    if(!open || buffer.empty())
```
z33ky, line 108: Shouldn't be just create an empty file if `buffer.empty()`? Is this an error?

z33ky (changes requested):

source/SH3/arc/vfile.cpp
Quaker762: Not so much an error, it just doesn't dump anything to the disk to save a few clock cycles. I didn't really see the point of creating an empty file, though I did initially have it in.

Quaker762:
@z33ky  I think I've just noticed a problem with how I have this set up. Some mesages may have more than one attribute, but if we change `currentAttr` during `Load()`, the old attribute will be lost. How do you think we should store all the attributes??

z33ky:
`currentAttr` is part of the other pull request #65. I'll answer there.
Could you add `msg.cpp` to the `source/CMakeLists.txt` as well?

Quaker762:
Oops, I didn't even notice that. Is this PR ready for a merge?

> Could you add msg.cpp to the source/CMakeLists.txt as well?

I have in the last pushed commit in #67

z33ky:
Doesn't look like it
```
$ git diff --name-only upstream/master quaker/dump
SILENT HILL 3 Redux.cbp
include/SH3/arc/vfile.hpp
include/SH3/system/msg/msg.hpp
source/SH3/arc/section.cpp
source/SH3/arc/types.cpp
source/SH3/arc/vfile.cpp
source/SH3/system/msg/msg.cpp
```

Quaker762:
Oops, I didn't realise you meant to the dump branch, I'll do that and rebasr asap.

z33ky (approved):

z33ky:
Well that is confusing. It seems you have some changes intended for #65 in here, as was also previously the case as seen by my `git diff` output before your rebase. You should remove those commits (including of course adding `msg.cpp` to `CMakeLists.txt`) from this pull request, especially since the message stuff is still subject to change.

Quaker762:
Okay, I really fucked that rebase up badly (I must've pushed a button accidentally...). I'm working on fixing it now...

Quaker762:
I'm deleting this branch and closing, I've fucked this up too hard.

PR #71, opened 2017-01-24
Added a function to dump files

Quaker762:
Added a function to dump files, so as to aid in reverse
engineering/inspecting files outside of their respective arc sections.

### DIFF
--- a/source/SH3/arc/vfile.cpp
+++ b/source/SH3/arc/vfile.cpp
@@ -13,6 +13,7 @@
 #include <SH3/system/log.hpp>
 #include <cstring>
 #include <cassert>
+#include <fstream>
 
 bool sh3_arc_vfile::Open(sh3_arc& mft, const std::string& filename)
 {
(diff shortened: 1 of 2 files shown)


Quaker762:
@z33ky This is the new PR, I really cooked the last one lol. Don't drink and git, I guess haha.

source/SH3/arc/vfile.cpp
```diff
@@ -100,3 +102,22 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
+void sh3_arc_vfile::Dump2Disk()
+{
+    if(!open || buffer.empty())
+    {
+        Log(LogLevel::WARN, "sh3_arc_vfile::Dump2Disk( ): Warning! Attempting to flush unopen or empty buffer to disk!");
+        return;
+    }
+
+    // Firstly, we need to convert the file path to just the filename.
+    std::stringstream ss(fname);
```
z33ky: unused. remove.
I think you can then also remove the `sstream` include.

source/SH3/arc/vfile.cpp
```diff
@@ -100,3 +102,22 @@ std::size_t sh3_arc_vfile::ReadData(void* destination, std::size_t len, read_err
...
+    std::size_t pos = fname.rfind('/');
+
+    std::ofstream out_file(fname.substr(pos + 1, fname.size()));
+    if(!out_file)
+        return;
+
+    out_file.write((const char*)&buffer[0], buffer.size());
```
z33ky, line 120: Doesn't `buffer.data()` (without cast) work for the first arguments?

z33ky (changes requested):

z33ky:
Yeah, rebasing is tricky in the presence of merges. Actually, checkout the documentation for `git-rebase` on `--preserve-merges`. Just a guess that's what happened though, didn't really check out what was going on in your branch there.

source/SH3/arc/vfile.cpp
Quaker762: Nah it doesn't, I get an `invalid conversion from unsigned char* to const char*` error

z33ky (approved):

PR #73, opened 2017-01-27
Initial Implementation of Shader Programs

Quaker762:
Can load, compile and link shaders from files. Currently untested (as we
don't have a GLContext working as of yet!)

@z33ky I'm going to add more to this over the next few days. We should probably start doing some GL stuff to test drawing/some other systems (like textures). Dis you want to try out fast-forwaring instead of merging this time?

### DIFF
--- a/source/SH3/system/glprogram.cpp
+++ b/source/SH3/system/glprogram.cpp
@@ -1,0 +1,136 @@
+/** @file
+ *
+ *  Implementation of glprogram.hpp
+ *
+ *  @copyright 2016  Palm Studios
+ *
+ *  @date
+ *
+ *  @author Jesse Buhagiar
+ */
+#include <SH3/system/glprogram.hpp>
+#include <SH3/system/log.hpp>
+
+#include <fstream>
+#include <vector>
+
+static const GLuint BAD_SHADER = 0;
+
+using namespace sh3_graphics;
+
+GLint sh3_glprogram::Load(const std::string& name)
+{
+    GLint status = 0;
+
+    programID = glCreateProgram();
+
+    // Generate our vertex and fragment shader
+    vertShader = (GLuint)Compile(GL_VERTEX_SHADER);
+    fragShader = (GLuint)Compile(GL_FRAGMENT_SHADER);
+
+    glAttachShader(programID, vertShader);
+    glAttachShader(programID, fragShader);
+    glLinkProgram(programID);
+
+    // Make sure our program was linked without any errors
+    glGetProgramiv(programID, GL_LINK_STATUS, (int*)&status);
+
+    if(status == GL_FALSE)
+    {
+        GLint logLength = 0;
+        std::vector<GLchar> errorLog;   // Error log
+        glGetShaderiv(programID, GL_INFO_LOG_LENGTH, &logLength);
+
+        errorLog.resize(logLength);
+        glGetProgramInfoLog(programID, logLength, &logLength, &errorLog[0]);
+        glDeleteShader(vertShader);
+        glDeleteShader(fragShader);
+
+        Log(LogLevel::ERROR, "%s\n%s", "glprogram::Load( ): Linking failed! Reason:", &errorLog[0]);
+
+        // TODO: Default shader fallback.
+    }
+
+    glDeleteShader(vertShader); // Blast these shaders into Oblivion!! They are no use to us anymore!
+    glDeleteShader(fragShader);
+}
+
+void sh3_glprogram::Bind()
+{
+    if(programID == 0)
+    {
+        Log(LogLevel::WARN, "sh3_glprogram::Bind( ): It seems this glprogram has not been compiled! It is impossible for it to be bound!");
+        return;
+    }
+
+    glUseProgram(programID);
+}
+
+void sh3_glprogram::Unbind()
+{
+    glUseProgram(0); // The 'correct' way to unbind a shader, but apparently undefined?!?!
+}
+
+// TODO: Unfuck this function! It looks like a pile of shit (and acts like one too)..
+GLuint sh3_glprogram::Compile(GLenum type)
+{
+    GLint   status; // Used to check compilation status
+    GLuint  id;     // ID of this shader
+
+    std::ifstream       source;     // Handle to the shader program source
+    std::string         ssource;    // Temporary source string to read into
+    std::string         fname;      // Name of the file we want to open
+    std::string         line;
+    std::vector<GLchar> errorLog;   // Error log
+
+    if(type == GL_VERTEX_SHADER)
+    {
+        id = glCreateShader(GL_VERTEX_SHADER); // Generate a shader ID
+        fname = "data/shaders/" + programName + ".vert";
+    }
+    else if(type == GL_FRAGMENT_SHADER)
+    {
+        id = glCreateShader(GL_FRAGMENT_SHADER); // Generate a shader ID
+        fname = "data/shaders/" + programName + ".frag";
+    }
+    else
+    {
+        die("glprogram::Compile( ): Only GL_VERTEX and GL_FRAGMENT shader programs are supported!");
+    }
+
+    source.open(fname);
+
+    if(!source.is_open())
+    {
+        //TODO: default shader fallback.
+        Log(LogLevel::ERROR, "sh3_glprogram::Compile( ): Unable to open a handle to %s!", fname.c_str());
+        return BAD_SHADER; // 0 is the 'bad shader' or rather one that is unbound.
+    }
+
+    while(!source.eof())
+    {
+        std::getline(source, line);
+        ssource += line + "\n";
+    }
+
+    const GLchar* csource = ssource.c_str();
+    glShaderSource(id, 1, &csource, 0);
+    glCompileShader(id);
+
+    glGetShaderiv(id, GL_COMPILE_STATUS, &status);
+    if(status == GL_FALSE) // Ooops, our shader compilation failed! D:
+    {
+        GLint logLength = 0;
+        std::string errmsg = "sh3_glprogram::Compile( ) error in " + fname;
+
+        glGetShaderiv(id, GL_INFO_LOG_LENGTH, &logLength);
+
+        errorLog.resize(logLength);
+        glGetShaderInfoLog(id, logLength, &logLength, &errorLog[0]);
+        glDeleteShader(id); // Probably pointless, considering we can't continue!
+
+        Log(LogLevel::ERROR, "%s\n%s", errmsg.c_str(), &errorLog[0]); // TODO: Default shader fallback
+    }
+
+    return id;
+}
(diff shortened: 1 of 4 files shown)


Quaker762:
Also see #74

Quaker762:
Okay I've fixed a weird segfault. This should be good to go. glShaderSource is a bizarre function...

EDIT: It isn't, I just realised my file IO on about line 93 of glprogram.cpp is REALLY bad. I'll fix it ASAP.

source/SH3/system/glprogram.cpp
```diff
@@ -0,0 +1,118 @@
...
+GLuint sh3_glprogram::Compile(std::uint32_t type)
+{
+    GLint   status; // Used to check compilation status
+    GLuint  id;     // ID of this shader
+
+    std::ifstream       source;     // Handle to the shader program source
+    std::string         fname;      // Name of the file we want to open
+    std::vector<GLchar> errorLog;   // Error log
+
+    if(type == GL_VERTEX_SHADER)
```
z33ky: There's plenty of code duplication in these branches.
For each shader type, only fname seems to be the difference.

source/SH3/system/glprogram.cpp
```diff
@@ -0,0 +1,118 @@
...
+        glGetShaderiv(id, GL_INFO_LOG_LENGTH, &logLength);
+
+        errorLog.resize(logLength);
+        glGetShaderInfoLog(id, logLength, &logLength, &errorLog[0]);
+        glDeleteShader(id); // Probably pointless, considering we can't continue!
+
+        die(&errorLog[0]);
+    }
+
+    source.close();
```
z33ky: No need for that, RAII will do that.

source/SH3/system/glprogram.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    if(status == GL_FALSE)
+    {
+        GLint logLength = 0;
+        std::vector<GLchar> errorLog;   // Error log
+        glGetShaderiv(programID, GL_INFO_LOG_LENGTH, &logLength);
+
+        errorLog.resize(logLength);
+        glGetProgramInfoLog(programID, logLength, &logLength, &errorLog[0]);
+        glDeleteProgram(vertShader);
```
z33ky: Should be `glDeleteShader` I think.

z33ky (changes requested):
Perhaps the `glDelete*` stuff should just be in the destructor of the class (even in the error case). This way resources will always be released when the `sh3_glprogram` is destroyed, be it because it couldn't be loaded or when it simply isn't needed anymore.

source/SH3/system/glprogram.cpp
```diff
@@ -51,21 +53,24 @@ GLint sh3_glprogram::Load(const std::string& name)
 
 void sh3_glprogram::Bind()
 {
-
+    if(programID != 0)
```
z33ky:
```
if(programID == 0)
{
    //warn
    return;
}
```
?

source/SH3/system/glprogram.cpp
```diff
@@ -51,21 +53,24 @@ GLint sh3_glprogram::Load(const std::string& name)
...
+        glUseProgram(programID);
 }
 
 void sh3_glprogram::Unbind()
 {
-
+    glUseProgram(0); // The 'correct' way to unbind a shader, but apparently undefined?!?!
```
z33ky: I don't have a huge issue with this, but as a fix we could use a dummy shader (and have a static class variable refer to it) that is set on `Unbind()`.
The dummy shader would either just render a solid color or an error message if we want to be fancy.

source/SH3/system/glprogram.cpp
```diff
@@ -75,11 +80,17 @@ GLuint sh3_glprogram::Compile(std::uint32_t type)
 
         source.open(fname);
 
+        if(!source.is_open())
+            die("sh3_glprogram::Compile( ): Unable to open a handle to %s!", fname.c_str());
+
         while(!source.eof())
```
z33ky: We could require the shader sources to have UNIX-style line endings ('\n'), in which case we can open the file in binary mode and use a `std::ostringstream` and read the whole file via `stringstream << source.rdbuf();`.
Reading the file like this is safer though.

source/SH3/system/glprogram.cpp
Quaker762: Yeah I'm getting on fixing this ASAP.

source/SH3/system/glprogram.cpp
Quaker762: Whoops haha

source/SH3/system/glprogram.cpp
Quaker762: Yeah that sounds like a good idea. Maybe even a checkered square like the Source engine haha.

source/SH3/system/glprogram.cpp
Quaker762: Ahh I see. I'm using Windows, so I think the problem is stemming from the stupid `\r\n` at the end of each line. 

Why is this method safer?

Quaker762:
Regarding `glDelete*()` calls, the OpenGL spec seems to recommend that glDeleteShader be called if compile or link fails. I've added `glDeleteShader()` and `glDeleteProgram()` calls to the destructor nonetheless.

source/SH3/system/glprogram.cpp
z33ky: Safer as in, doesn't lead to errors when using different line endings.
You should be able to switch to different line endings in your editor.

source/SH3/system/glprogram.cpp
```diff
@@ -39,7 +41,7 @@ GLint sh3_glprogram::Load(const std::string& name)
 
         errorLog.resize(logLength);
         glGetProgramInfoLog(programID, logLength, &logLength, &errorLog[0]);
-        glDeleteProgram(vertShader);
+        glDeleteShader(vertShader);
         glDeleteShader(fragShader);
 
         die(&errorLog[0]);
```
z33ky: When the string contains something looking like a format specifier, things will go wrong. Use `die("%s", ...)` instead.

z33ky (changes requested):

z33ky:
The `glDestroy*` functions will still be called when the destructor is run. Would this be insufficient?
Also, now it will be called twice when a shader fails to compile, once in `Load()` and once in `sh3_glprogram`.

Quaker762:
> The glDestroy* functions will still be called when the destructor is run. Would this be insufficient?

If anything, they should be destroyed right after the link is performed, as they are no longer needed and are just a waste of memory (as I have read, but it seems everyone has their own opinion on this stuff haha), though with the amount most people have in their system, this is probably a non-issue (though if we preload stuff it may be).

> Also, now it will be called twice when a shader fails to compile, once in Load() and once in sh3_glprogram

Will the constructor still be run if a call to `die()` is made??? I thought that was only the case if they're static/global objects. 
Also, now that I think about it, should we actually die if it fails to compile, or just replace the shader with a default one as you've suggested somewhere in the comment chain?

source/SH3/system/glprogram.cpp
Quaker762: Completely missed that, it's fixed now.

z33ky:
I though that `die()` was just a sort of "error-handling to be implemented later" in most cases.
Like stuff not being found in an arc doesn't seem to be a truly fatal error in many cases. At worst it should just abort loading the current scene, not the whole program.

Quaker762:
> I though that die() was just a sort of "error-handling to be implemented later" in most cases.

True, we should probably just keep them to aid with debugging shaders at this point. 

> Like stuff not being found in an arc doesn't seem to be a truly fatal error in many cases

Wouldn't that imply the file is corrupt or damaged in some way though?

Also, do you want me to move the `glDeleteShader()` calls to straight after the program has been linked, and leave just `glDeleteProgram()` in the destructor?

z33ky:
> Wouldn't that imply the file is corrupt or damaged in some way though?

That or driver issues probably.
I think we should make a best effort. If something goes wrong we should definitely log, but just fallback to the dummy shader. It probably won't make for a good playing experience, but if multiple shader fail for instance, then you get all errors in one run, instead of a cycle of run, die, fix shader until all are fixed.

z33ky:
> Also, do you want me to move the `glDeleteShader()` calls to straight after the program has been linked, and leave just `glDeleteProgram()` in the destructor?

Yeah, that seems good. Although now `glDeleteShader()` is missing for the successful case.
The spec is okay with deleting the shader while still attached; It will then be deleted when the shader is detached. So we should just always do that after the shader was attached.
The program will also automatically detach the shaders, so we don't need to do that manually.

Quaker762:
> I think we should make a best effort. If something goes wrong we should definitely log, but just fallback to the dummy shader. It probably won't make for a good playing experience, but if multiple shader fail for instance, then you get all errors in one run, instead of a cycle of run, die, fix shader until all are fixed.

Good point, I'll get rid of all of the calls to `die()` and replace them with `Log()` with the `ERROR` status.

> Yeah, that seems good. Although now glDeleteShader() is missing for the successful case.

Bloody hell, more of my 1AM programming... I think I should make myself a coffee before I attempt to program (that or crack open a beer). I've fixed it now.

> The spec is okay with deleting the shader while still attached

I had no clue this was the case. Everything I read 'recommended' a detach before delete. Very interesting. I've removed all calls to detach.

source/SH3/system/glprogram.cpp
```diff
@@ -93,7 +95,7 @@ GLuint sh3_glprogram::Compile(GLenum type)
     source.open(fname);
 
     if(!source.is_open())
-        die("sh3_glprogram::Compile( ): Unable to open a handle to %s!", fname.c_str());
+        Log(LogLevel::ERROR, "sh3_glprogram::Compile( ): Unable to open a handle to %s!", fname.c_str());
```
z33ky: I think we can just `return` at this point.
`//TODO: default shader fallback.`

z33ky:
You can also squash the commits. I don't see a good reason to keep them separate.

source/SH3/system/glprogram.cpp
```diff
@@ -0,0 +1,132 @@
...
+    if(type == GL_VERTEX_SHADER)
+    {
+        id = glCreateShader(GL_VERTEX_SHADER); // Generate a shader ID
+        fname = "data/shaders/" + programName + ".vert";
+    }
+    else if(type == GL_FRAGMENT_SHADER)
+    {
+        id = glCreateShader(GL_FRAGMENT_SHADER); // Generate a shader ID
+        fname = "data/shaders/" + programName + ".frag";
+    }
```
z33ky, line 95:
```
else
    die()
```

source/SH3/system/glprogram.cpp
```diff
@@ -0,0 +1,132 @@
...
+ *
+ *  @author Jesse Buhagiar
+ */
+#include <SH3/system/glprogram.hpp>
+#include <SH3/system/log.hpp>
+
+#include <fstream>
+#include <vector>
+
+#define BAD_SHADER 0
```
z33ky: This should be a `static const GLuint` IMO.

z33ky (approved):

PR #75, opened 2017-01-31
Added getter to 'fsize' in vfile.hpp

Quaker762:
Added a getter to vfile, as it was not possible to access the file size, therby making it impossible to read a file in full.


z33ky:
This method can be `const`.

Quaker762:
Done, I completely forgot you could do this haha.

Issue #72, opened 2017-01-24
Merging workflow


z33ky:
Currently we use a merge-heavy workflow, which of course results in plenty merge commits. Many of those are fast forwardable, i.e. "unnecessary" merges.
Having merges even when fast forwardable commits are integrated preserves the branch structure. On the other hand, it litters the log with merge commits.
So to get to the point, I am wondering whether we want to continue with a merge-heavy workflow, or if we want to use fast forwarding (i.e. avoiding merge commits) when possible. Note that in pull requests you have the option of using rebase instead of merge to integrate the changes, which avoids the merge commit.
I can also `--force` push a modified, linear history for the existing history, which would require everyone else to do a `git reset --hard` on their master branches (assuming no additional commits were made in private; a `git pull --rebase` might throw a tantrum about the changed history) though - but only as a one-time thing.

Quaker762:
I'll leave this decision up to you. I don't really know all that much about git/version control (unlike you), so make whatever decision is best (but make sure to give me instructions haha).

z33ky:
There are a couple of issues with how git handles conflict resolution when rebasing, but I'm a bigger fan of the linear history. I'll get on it soon (today or tomorrow).

z33ky:
> I'll get on it soon (today or tomorrow).

Hhheh.

So I created the branch `master-linear`.
Try rebasing the branches you need onto `master` first, so that they're up-to-date, and then on that `master-linear` branch.
If you got no problems there I'll rename `master` to `old-master` (which we can remove after a while) and `master-linear` to `master`.

Quaker762:
> Try rebasing the branches you need onto master first, so that they're up-to-date, and then on that master-linear branch.

How would I go about doing this? I'll probably screw my fork if I attempt this myself (I have literally no clue lol). Is this for stuff I'm currently working on??

> If you got no problems there I'll rename master to old-master (which we can remove after a while) and master-linear to master

Alright sure thing. Also could you update `Readme.MD` with a section on how to make changed?

z33ky:
> How would I go about doing this?

First let's bring master up-to-date with `git checkout master && git pull upstream`. Then for each other branch, do `git rebase master`, fix the conflicts if you get any and finally `git rebase upstream/master-linear` (assuming you named the Palm-Studios remote `upstream`).
You can set your master branch to be master-linear via `git reset --hard upstream/master`.

[...]

PR #76, opened 2017-02-01
Reworked Base texture class

Quaker762:
Textures can now be bound for operations by OpenGL, as well as the
ability to be loaded by the new Virtual File System we have. Textures
are stored as their pixel array in a vector.

Currently, 8-bit textures aren't supported because for reasons that
elude me, they are palletted, and I have no clue as to where the
pallete is (or what data it contains).

The header was also fixed up, as it was missing a 16-bit unused value,
which was distorting the number of bytes passed to the Virtual File
System to load the texture.

### DIFF
--- a/include/SH3/graphics/texture.hpp
+++ b/include/SH3/graphics/texture.hpp
@@ -97,8 +97,6 @@ namespace sh3_graphics
 
     private:
         GLuint tex;                         /**< ID representing this texture */
-
-        std::vector<std::uint8_t> data;     /**< Pixel data of this texture (with the header stripped) */
     };
 }
 
--- a/include/SH3/graphics/texture.hpp
+++ b/include/SH3/graphics/texture.hpp
@@ -8,21 +8,21 @@
             all of the file types, including all of the textures, meshes and even the motion capture
             skeletal animation that Konami captured. Thanks mate!
  *
- *  @date 2/1/2017
+ *  @note 8-bit textures are currently unsupported, as they are palletted. See <a href=https://www.reddit.com/r/REGames/comments/5kp62x/help_reverse_engineering_silent_hill_3_texture/?st=iyn04asn&sh=92f998d9>here</a>
+ *
+ *  @date 2-1-2017
  *
  *  @author Jesse Buhagiar
  */
 #ifndef SH3_TEXTURE_HPP_INCLUDED
 #define SH3_TEXTURE_HPP_INCLUDED
 
-#include <SH3/arc/types.hpp>
+#include "SH3/arc/types.hpp"
+#include "SH3/arc/vfile.hpp"
 
-#include <cstdint>
-#include <limits>
-#include <vector>
+#include <GL/glew.h>
 #include <GL/gl.h>
 
-#include "SH3/error.hpp"
 
 namespace sh3_graphics
 {
@@ -35,29 +35,29 @@ namespace sh3_graphics
      *
      * Both batch and individual texture.
      */
-    typedef struct
+    struct sh3_texture_header
     {
-        std::uint32_t batchHeaderMarker;   /**< This should be 0xFFFFFFFF to mark a header chunk */
-        std::uint8_t  unused1[4];          /**< There are a lot of unused DWORDs, I assume to align everything nicely */
-        std::uint32_t batchHeaderSize;     /**< Size of the first part of the whole header */
-        std::uint32_t batchSize;           /**< = res * res * (bpp/8) * #tex + 128 * #tex */
-        std::uint8_t  unused2[4];          /**< Unused **/
-        std::uint32_t numBatchedTextures;  /**< Number of textures in this texture file */
-        std::uint8_t  unused3[8];          /**< Unused **/
-        std::uint32_t texHeaderSegMarker;  /**< Secondary texture marker. This signifies the start of the texture information header. This is also 0xFFFFFFFF */
-        std::uint8_t  unused4[4];          /**< Unused **/
-        std::uint16_t texWidth;            /**< The width of this texture */
-        std::uint16_t texHeight;           /**< The height of this texture */
-        std::uint8_t  bpp;                 /**< Number of bits per pixel of this texture. NOTE: 8bpp is believed to be paletted! */
-        std::uint8_t  dataOffset;          /**< #bytes from texHeaderSize+16 to 128 (0 filled) */
-        std::uint8_t  unused5[4];          /**< Unused **/
-        std::uint32_t texSize;             /**< The size of this texture in bytes (w * h * [bpp/8]) */
-        std::uint32_t texHeaderSize;       /**< = texSize + texHeaderSize + 16 + endFilleSize */
-        std::uint8_t  unused6[4];          /**< Unused **/
-        std::uint32_t unknown1;            /**< Completely unknown, probably unimportant for now */
-        std::uint32_t unknown2;            /**< Usually 1! */
-        std::uint32_t unused7[15];         /**< Unused **/
-    } sh3_texture_header;
+        std::uint32_t batchHeaderMarker;    /**< This should be 0xFFFFFFFF to mark a header chunk */
+        std::uint8_t  unused1[4];           /**< There are a lot of unused DWORDs, I assume to align everything nicely */
+        std::uint32_t batchHeaderSize;      /**< Size of the first part of the whole header */
+        std::uint32_t batchSize;            /**< = res * res * (bpp/8) * #tex + 128 * #tex */
+        std::uint8_t  unused2[4];           /**< Unused **/
+        std::uint32_t numBatchedTextures;   /**< Number of textures in this texture file */
+        std::uint8_t  unused3[8];           /**< Unused **/
+        std::uint32_t texHeaderSegMarker;   /**< Secondary texture marker. This signifies the start of the texture information header. This is also 0xFFFFFFFF */
+        std::uint8_t  unused4[4];           /**< Unused **/
+        std::uint16_t texWidth;             /**< The width of this texture */
+        std::uint16_t texHeight;            /**< The height of this texture */
+        std::uint8_t  bpp;                  /**< Number of bits per pixel of this texture. NOTE: 8bpp is believed to be paletted! */
+        std::uint8_t  dataOffset;           /**< #bytes from texHeaderSize+16 to 128 (0 filled) */
+        std::uint8_t  unused5[2];           /**< Unused **/
+        std::uint32_t texSize;              /**< The size of this texture in bytes (w * h * [bpp/8]) */
+        std::uint32_t texFileSize;          /**< = texSize + texHeaderSize + 16 + endFilleSize */
+        std::uint8_t  unused6[4];           /**< Unused **/
+        std::uint32_t unknown1;             /**< Completely unknown, probably unimportant for now */
+        std::uint32_t unknown2;             /**< Usually 1! */
+        std::uint32_t unused7[15];          /**< Unused **/
+    };
 
     /**@}*/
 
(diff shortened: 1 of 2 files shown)


source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,57 @@
...
+    sh3_arc_vfile file(mft, filename);
+    sh3_arc_vfile::read_error e;
+
+    int ret = file.ReadData(&header, sizeof(sh3_texture_header), e); // Read in the header
+
+    if(ret != sizeof(sh3_texture_header))
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Header only read %d bytes of data! It should have been %d", ret, sizeof(sh3_texture_header));
+
+    if(header.bpp == 8)
+        die("8-bit textures are currently unsupported!");
```
z33ky: I think `die()` is a little drastic. We should warn here and get things rendering, even if it's messed up at the moment.
We could also check for other bpps we don't expect.

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,57 @@
...
-};
-
-texture::load_error sh3_graphics::texture::Load(const std::string& filename, sh3_arc& arc)
-{
-    load_error error;
+    sh3_texture_header header;
+    sh3_arc_vfile file(mft, filename);
+    sh3_arc_vfile::read_error e;
+
+    int ret = file.ReadData(&header, sizeof(sh3_texture_header), e); // Read in the header
```
z33ky: Should be `std::size_t`.
I"m also in favor of using `sizeof(header)` (generally `sizeof(variable)` where applicable) because the intention here is to read the size of the variable `header`, which just happens to be a `sh3_texture_header` here.

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,57 @@
...
+
+    Log(LogLevel::INFO, "Texture info:\n\ttexWidth:%d\n\ttexHeight:%d\n\ttexSize:%d\n\tbpp:%d", header.texWidth, header.texHeight, header.texSize, header.bpp);
+
+    file.Seek(sizeof(sh3_texture_header) + header.dataOffset, std::ios_base::beg);
 
-    if(arc.LoadFile(filename, pixbuff) == ARC_FILE_NOT_FOUND)
-    {
-        die("sh3_graphics::texture::Load( ): Unable to find file %s!", filename.c_str());
-    }
+    data.resize(header.texSize);
```
z33ky: We could check that `texSize == texWidth * texHeight * bpp`. That should apply, right?
If it doesn't we should make sure that there's no out-of-bounds read from `data`.

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,57 @@
...
 
-    error.set_error(load_result::END_OF_FILE);
+    if(ret != header.texSize)
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Only read %d out of %d bytes!", ret, header.texSize);
 
-    return error;
+    // Now actually create a logical texture with OpenGL on the gpu
+    glGenTextures(1, &tex);
+
+    glBindTexture(GL_TEXTURE_2D, tex);
```
z33ky: This could use `Bind()` and later `Unbind()`. Though apart from being shorter I don't see much benefit.

z33ky (changes requested):
The texture also needs to be deleted at some point.

Have you considered bindless textures? I don't know if that is something we do want to use, it's just something that I've heard about and lingers in my mind.

Quaker762:
> The texture also needs to be deleted at some point.

[Lmao](https://cdn.meme.am/cache/instances/folder610/500x/75266610.jpg)

> Have you considered bindless textures? I don't know if that is something we do want to use, it's just something that I've heard about and lingers in my mind.

I had never even heard of these until I googled them. It seems a little bit weird though (and the Khronos site even mentions it's a bit complex [here](https://www.khronos.org/opengl/wiki/Bindless_Texture))
I feel like binding makes a little bit more sense. You bind the texture you want to do something on, perform it, and then let it free. They'd be cool to look into and implement though, as long as they're not overly complex haha.

source/SH3/graphics/texture.cpp
Quaker762: > I"m also in favor of using sizeof(header) (generally sizeof(variable)

I'll do this from now on then. I think the whole `sizeof(typename)` has stuck from not only books I've read, but other peoples' code I've looked through.

source/SH3/graphics/texture.cpp
Quaker762: > I think die() is a little drastic

I thought so too, but the screen will end up looking like vomit [as you can see here](http://imgur.com/a/1lxeN). I'll remove it and log for now though.

source/SH3/graphics/texture.cpp
Quaker762: As in an assertion??

source/SH3/graphics/texture.cpp
Quaker762: > This could use Bind() and later Unbind()

I have this weird habit of not using class functions inside other class functions... Strange I know..

Quaker762:
Also I just realised I should add texture units into bind, so we can bind to different texture units for different textures [as it explained here](https://www.opengl.org/discussion_boards/showthread.php/173263-What-is-a-texture-unit)

source/SH3/graphics/texture.cpp
z33ky: Assertions should only be present in debug mode, so I think this should be a regular check & log.

include/SH3/graphics/texture.hpp
```diff
@@ -69,47 +69,36 @@ namespace sh3_graphics
...
-         * @param filename - Virtual file path to file.
-         * @param arc - Reference to sh3_arc (arc.arc)
-         *
-         * @return @ref load_result indicating whether the read failed or not
+         *  Bind this texture for use with any draw calls
          *
+         *  @param textureUnit The texture unit we want to bind this texture to
          */
-        load_error Load(const std::string& filename, sh3_arc& arc);
+         void Bind(GLenum textureUnit);
```
z33ky: 1 space too many

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,62 @@
...
+    sh3_arc_vfile file(mft, filename);
+    sh3_arc_vfile::read_error e;
+
+    int ret = file.ReadData(&header, sizeof(sh3_texture_header), e); // Read in the header
+
+    if(ret != sizeof(sh3_texture_header))
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Header only read %d bytes of data! It should have been %d", ret, sizeof(sh3_texture_header));
+
+    if(header.bpp == 8)
+        die("8-bit textures are currently unsupported!");
```
z33ky: `die()` and `Log()`?

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,62 @@
...
-{
-    load_error error;
+    sh3_texture_header header;
+    sh3_arc_vfile file(mft, filename);
+    sh3_arc_vfile::read_error e;
+
+    int ret = file.ReadData(&header, sizeof(sh3_texture_header), e); // Read in the header
+
+    if(ret != sizeof(sh3_texture_header))
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Header only read %d bytes of data! It should have been %d", ret, sizeof(sh3_texture_header));
```
z33ky: We may want to fallback on a dummy texture when this happens, since it would be very cumbersome to ensure we don't read uninitialized struct fields.
For now this would be a good place to `die()`, though add a `//TODO` that we want to use a fallback texture in the future.

z33ky (changes requested):

source/SH3/graphics/texture.cpp
Quaker762: Crap I forgot to change this...

z33ky:
> I feel like binding makes a little bit more sense. You bind the texture you want to do something on, perform it, and then let it free. They'd be cool to look into and implement though, as long as they're not overly complex haha.

I find the stateful implementation weirder than the stateless one. The stateless version is OOP-like `do_stuff(thingToDoStuffOn, parameters...);` whereas the stateful has this verbose pattern of
```
make_thing_active(thingToDoStuffOn);
do_stuff(parameters...);
//sometimes: make_thing_inactive(thingToDoStuffOn);
```
Even worse if you made some function-calls between `make_thing_active` and `do_stuff`, because you have to hope that `thingToDoStuffOn` is still active, or you need to re-activate it just in case.
At least this is my understanding of the concept, without having read anything about the actual API.

Quaker762:
> I find the stateful implementation weirder than the stateless one

Aren't programming paradigms and semantics fun haha.

> Even worse if you made some function-calls between make_thing_active and do_stuff, because you have to hope that thingToDoStuffOn is still active, or you need to re-activate it just in case.

Very true, though the chance of that happening are pretty slim with the amount we're reviewing each others code. Though I do see the problem in this. OpenGL wont shit itself if this happens, so your calls just silently fail or even worse corrupt something.

> At least this is my understanding of the concept, without having read anything about the actual API.

It seems like it. The only other issue I have is this is an ARB extension, which I assume most modern graphics cards support, but the chances of this being run on someone's grandma's PC is pretty high haha.

z33ky:
> The only other issue I have is this is an ARB extension, which I assume most modern graphics cards support, but the chances of this being run on someone's grandma's PC is pretty high haha.

I might have an idea how we can support both in a uniform API. I'll try that when this PR is ~merged~ rebased ;).

Quaker762:
> I'll try that when this PR is merged rebased ;)

What do you need me to rebase?? Or is this the new workflow method you were talking about in #72 haha?

z33ky:
> Or is this the new workflow method you were talking about in #72 haha?

Yeah. The rebase will happen automatically if there's no conflicts.

Quaker762:
> The rebase will happen automatically if there's no conflicts.

How the hell does that work!?!?!

z33ky:
See https://github.com/Palm-Studios/sh3redux/pull/76#pullrequestreview-19780384

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,62 @@
...
+    if(ret != header.texSize)
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Only read %d out of %d bytes!", ret, header.texSize);
 
-    return error;
+    // Now actually create a logical texture with OpenGL on the gpu
+    glGenTextures(1, &tex);
+
+    glBindTexture(GL_TEXTURE_2D, tex);
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, header.texWidth, header.texHeight, 0, GL_BGRA, GL_UNSIGNED_BYTE, &data[0]);
+    glBindTexture(GL_TEXTURE_2D, 0); // Unbind this texture
```
z33ky, line 57: Should this restore the previously bound texture (can be obtained via `glGetIntegerv(GL_TEXTURE_BINDING_2D, &prevTex);` I think)?

source/SH3/graphics/texture.cpp
z33ky: This is still missing.

source/SH3/graphics/texture.cpp
z33ky: This is still missing.

z33ky (changes requested):

source/SH3/graphics/texture.cpp
Quaker762: I don't quite understand why we'd want to bind the previous texture? This only changes a few parameters for this texture during a load. When the actual game is running, textures are bound and unbound when they're drawn (or rather, bound to different texture units and drawn) .

source/SH3/graphics/texture.cpp
Quaker762: Fixed

source/SH3/graphics/texture.cpp
z33ky: You mean they'll be (re-)bound per frame? That won't be an issue then.

Quaker762: Yeah they have to be to be drawn. Basically (how I'm thinking it'll work), is each mesh has a texture (*.tex). Everything is loaded before the scene is rendered. During the render loop, we transverse a list containing each mesh. Its VBO and texture is bound and drawn etc. 

When you generate a texture I'm pretty sure it's recommended you unbind it. Otherwise you could add a bunch of stuff to another texture without realising it.

source/SH3/graphics/texture.cpp
```diff
@@ -4,44 +4,69 @@
...
+        Log(LogLevel::WARN, "WARNING: 8-bit textures are currently unsupported! They WILL look like nothing but pixel garbage.");
+
+    Log(LogLevel::INFO, "Texture info:\n\ttexWidth:%d\n\ttexHeight:%d\n\ttexSize:%d\n\tbpp:%d", header.texWidth, header.texHeight, header.texSize, header.bpp);
+
+    file.Seek(sizeof(header) + header.dataOffset, std::ios_base::beg);
 
-    if(arc.LoadFile(filename, pixbuff) == ARC_FILE_NOT_FOUND)
+    data.resize(header.texSize);
+
+    if(header.texSize == header.texWidth * header.texHeight * header.bpp)
```
z33ky: Do an early return, i.e. instead of
```
if(x)
{
    A;
}
else
{
    B;
    return;
}
```
do
```
if(!x)
{
    B;
    return;
}
A;
````

Quaker762: Done

z33ky (approved):

PR #78, opened 2017-02-02
Add system/glprogram.cpp to CMakeLists.txt

z33ky:
What prevents you from using CMake to generate the Code::Blocks project? :/

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -22,6 +22,7 @@ add_executable("${CMAKE_PROJECT_NAME}"
 	"SH3/system/assert.cpp"
 	"SH3/system/config.cpp"
 	"SH3/system/glcontext.cpp"
+	"SH3/system/glprogram.cpp"
 	"SH3/system/log.cpp"
 	"SH3/system/window.cpp"
 	


Quaker762:
Mainly Windows... I have all my dependencies in a 'lib' folder. CMake tends to screw up a little bit on Windows because of dependency paths etc, it's pretty bad. That and I like having both the i686 and x86_64 targets, because I've worked on stuff before where one has crashed and the other hasn't (though this may have been whacky compiler output due to optimisations).

I'll try and get CMake working properly on my PC, but in the future I'll remember to add new files to the `CMakeLists.txt` file haha

z33ky:
> I have all my dependencies in a 'lib' folder. CMake tends to screw up a little bit on Windows because of dependency paths etc, it's pretty bad.

If it does it's probably my fault.
You should be able to set paths for all the dependencies.

> That and I like having both the i686 and x86_64 targets, because I've worked on stuff before where one has crashed and the other hasn't (though this may have been whacky compiler output due to optimisations).

This would be useful for testing. While CMake does not support this natively we can add architecture-specific `CMAKE_CONFIGURATION_TYPES` that pass `-m32`.

Quaker762:
> If it does it's probably my fault.

Nah, it's just the way Windows handles deps unfortunately. I could all them all to the mingw lib folder but then you're in for a whole heap of trouble haha.

> You should be able to set paths for all the dependencies.

I'll look into doing this, I'm not too familiar with CMake

> This would be useful for testing

Which is why I still like using my .cbp project haha.

> While CMake does not support this natively we can add architecture-specific CMAKE_CONFIGURATION_TYPES that pass -m32

That's actually a good idea. When we get this going I'll fully switch over to using CMake.

PR #81, opened 2017-02-06
Initial bmp loader

Quaker762:
Created a small bitmap loader to deal with some of the bitmaps shipped
with sh3, as well as our default 'error' texture for textures that are
loaded/buond incorrectly. Currently just loads the texture, as there is
no shader to draw any of this, so Draw() is currently unimplemented.

I'll probably add more to this as I go. We should probably start getting some shaders done so we can actually draw stuff to the screen haha.

EDIT: I didn't check if the Konami logo bitmaps are actually 8-bit, so a fair chunk of this might get axed.. Pretty good exercise for those 8-bit textures we'll eventually have to implement though (if we can find the bloody colour palette).


include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /** @defgroup graphics-headers Graphics headers
+     *  @{
+     */
+
+     #pragma pack(push, 1) // Stop the compiler from giving us stupid naturally aligned boundaries (__attribute__((packed)) didn't work...)
+
+    /**
+     * Bitmap file information header.
+     */
+    typedef struct
```
z33ky: `typedef struct {} name;` is a C pattern. In C++ you just do `struct name {};`.

include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     *  Describes a Microsoft bitmap file, which can be loaded and sent to the graphics card.
+     *  Note that all bmp files are stored on disk PHYSICALLY, therefore do not require the use
+     *  of the virtual filesystem.
+     */
+    struct msbmp final
+    {
+    public:
+        msbmp(const std::string& path){Load(path);}
+        ~msbmp(){};
```
z33ky: Extra unneeded `;`.
Actually, the destructor doesn't need to be defined, it will be implicitly. I don't mind it being around either though.

source/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ add_executable("${CMAKE_PROJECT_NAME}"
 	"SH3/arc/vfile.cpp"
 	
 	"SH3/graphics/texture.cpp"
+	"SH3/graphics/msbmp.cpp"
```
z33ky, line 21: Nice :)

include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+        /**
+         *  Load all image data
+         *
+         *  @param path Physical path to the file (bmps are NOT contained in an arc section)
+         */
+        int Load(const std::string& path);
```
z33ky: The return parameter should be documented.
It's an error, for which I'd prefer an `error` (from `error.hpp`).

z33ky (changes requested):
I've just taken a look at the interface in this review. I'll look at the implementation file later.

include/SH3/graphics/msbmp.hpp
Quaker762: Interesting, I didn't know this was possible..

include/SH3/graphics/msbmp.hpp
Quaker762: I'll remove it. It's a bit of a habit I have when I'm in the header file.  I usually add the destructor in case we need to add anything. It just makes the class look more complete haha.

include/SH3/graphics/msbmp.hpp
Quaker762: Sure thing

source/CMakeLists.txt
Quaker762: HAHAHA I finally remembered.

include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include <GL/gl.h>
+
+namespace sh3_graphics
+{
+    /** @defgroup graphics-headers Graphics headers
+     *  @{
+     */
+
+     #pragma pack(push, 1) // Stop the compiler from giving us stupid naturally aligned boundaries (__attribute__((packed)) didn't work...)
```
z33ky, line 32: Have you tried `struct __attribute__((packed)) my_struct { ... };`?
You can check that there's no padding via `-Wpadded` BTW (enabled with CMake if `BUILD_EXTRA_WARNINGS` is set to `Many`).

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+    file.open(path, std::ios::binary);
+
+    if(!file.is_open())
+    {
+        // If this happens we're in a bit of trouble, because we can't load the default texture..... We could crash here.
+        Log(LogLevel::ERROR, "msbmp::Load( ): Unable to open a handle to %s!", path.c_str());
+        return res;
+    }
+
+    file.read((char*)&fheader, sizeof(fheader)); // Read in the file header
```
z33ky: `-Wold-style-cast`, use `reinterpret_cast`.
This should also appear in other places in this file.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+        file.seekg(sizeof(fheader) + sizeof(iheader), std::ios_base::beg); // Seek to the palette
+        file.read((char*)&palette[0], iheader.palette_size);
+
+        // Slightly expensive, but we now copy each pixel into the data buffer by getting it's index in the palette.
+        // Seeing as this is done once, I doubt it matters too much, especially if we do this during load time
+        // and keep the image alive for the whole of the scene duration (there are 3 bmps total in SH3, not including
+        // our default 'error' texture)
+        for(int i = 0; i < width * height; i += 3)
+        {
+            std::uint32_t   colour; // 4-byte colour, stored in BGRA (where A is unused)
```
z33ky: I'd prefer American spelling.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+            std::uint8_t    index;  // Image byte, which is the index into the colour palette
+            file.read((char*)&index, sizeof(std::uint8_t));
+
+            colour = palette[index];
+            data[i] = (colour >> 8) & 0xFF; // Extract Red
+            data[i + 1] = (colour >> 16) & 0xFF; // Extract Green
```
z33ky: Since `palette` consists of `std::uint8_t`'s, these are always going to be `0`.
`palette` should probably be an vector of `std::array<std::uint8_t, 3>`, or a `struct rgb_color`.
`data` should get the same treatment.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        // Slightly expensive, but we now copy each pixel into the data buffer by getting it's index in the palette.
+        // Seeing as this is done once, I doubt it matters too much, especially if we do this during load time
+        // and keep the image alive for the whole of the scene duration (there are 3 bmps total in SH3, not including
+        // our default 'error' texture)
+        for(int i = 0; i < width * height; i += 3)
+        {
+            std::uint32_t   colour; // 4-byte colour, stored in BGRA (where A is unused)
+            std::uint8_t    index;  // Image byte, which is the index into the colour palette
+            file.read((char*)&index, sizeof(std::uint8_t));
```
z33ky: `sizeof(index)` :3

include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,130 @@
...
+         */
+        std::int32_t GetWidth() const {return width;}
+
+        /**
+         *  Get the height of this texture
+         */
+        std::int32_t GetHeight() const {return height;}
+
+    private:
+        std::vector<std::uint8_t> palette;  /**< Vector containing colour data for this image */
```
z33ky: This is only needed during `Load()`, right? This could be a local variable then.

include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,130 @@
...
+        std::vector<std::uint8_t> data;     /**< Vector containing image data or palette offsets */
+
+        std::int32_t width;     /**< Width of this texture in pixels */
+        std::int32_t height;    /**< Height of this texture in pixels */
+        std::uint16_t bpp;      /**< Bitmap bit depth */
```
z33ky: Same as `palette`, this could be a local variable I think.

include/SH3/graphics/msbmp.hpp
```diff
@@ -0,0 +1,130 @@
...
+         *  Get the height of this texture
+         */
+        std::int32_t GetHeight() const {return height;}
+
+    private:
+        std::vector<std::uint8_t> palette;  /**< Vector containing colour data for this image */
+        std::vector<std::uint8_t> data;     /**< Vector containing image data or palette offsets */
+
+        std::int32_t width;     /**< Width of this texture in pixels */
+        std::int32_t height;    /**< Height of this texture in pixels */
```
z33ky: We could look into using a 2D container, but for now I'm fine with this.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+    else if(bpp == 24)
+    {
+        file.seekg(sizeof(fheader) + sizeof(iheader) + iheader.palette_size, std::ios_base::beg); // Seek to the image data
+        file.read((char*)&data[0], width * height * 3); // Read in the image data
+    }
+    else
+    {
+        Log(LogLevel::WARN, "msbmp::Load( ): Bad pixel depth!");
```
z33ky: Also print which pixel depth we got.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+            colour = palette[index];
+            data[i] = (colour >> 8) & 0xFF; // Extract Red
+            data[i + 1] = (colour >> 16) & 0xFF; // Extract Green
+            data[i + 2] = (colour >> 24) & 0xFF; // Extract Blue
+        }
+    }
+    else if(bpp == 24)
+    {
+        file.seekg(sizeof(fheader) + sizeof(iheader) + iheader.palette_size, std::ios_base::beg); // Seek to the image data
+        file.read((char*)&data[0], width * height * 3); // Read in the image data
```
z33ky: We already computed `width * height * 3`: `data.size()`.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+    {
+        palette.resize(iheader.palette_size);
+        file.seekg(sizeof(fheader) + sizeof(iheader), std::ios_base::beg); // Seek to the palette
+        file.read((char*)&palette[0], iheader.palette_size);
+
+        // Slightly expensive, but we now copy each pixel into the data buffer by getting it's index in the palette.
+        // Seeing as this is done once, I doubt it matters too much, especially if we do this during load time
+        // and keep the image alive for the whole of the scene duration (there are 3 bmps total in SH3, not including
+        // our default 'error' texture)
+        for(int i = 0; i < width * height; i += 3)
```
z33ky: I think this should go to `width * height * 3`, aka `data.size()`.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,100 @@
...
+    glGenTextures(1, &tex); // Bind texture for use
+    glBindTexture(GL_TEXTURE_2D, tex);
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, width, height, 0, GL_RGB, GL_UNSIGNED_BYTE, &data[0]);
+
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_REPEAT);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_REPEAT);
+
+    glBindTexture(GL_TEXTURE_2D, 0); // Unbind texture
```
z33ky, line 111: Same concern as #76 (restore the previously bound texture?).

z33ky (changes requested):

include/SH3/graphics/msbmp.hpp
Quaker762: The compiler must be spitting out bad code??? `__attribute__((packed))` usually always works for me, but this time it isn't... Perhaps it's a bug in my version of GCC??

http://imgur.com/a/TkpzV

include/SH3/graphics/msbmp.hpp
Quaker762: Yeah true. I'll move it.

include/SH3/graphics/msbmp.hpp
Quaker762: I like having them separated as a like to imagine a texture object an actual texture (even though it's technically not)

include/SH3/graphics/msbmp.hpp
Quaker762: Good point. I didn't even think that this won't be accessed anywhere else.

source/SH3/graphics/msbmp.cpp
Quaker762: Oops

EDIT: I just realised we don't need to do this (I think). If anything, it should be i++. Each byte in the data area is actually just an index into the colour table (of which each entry is 32-bits wide).

source/SH3/graphics/msbmp.cpp
Quaker762: Hahah eww. It's going to take a lot to get me to change that ;)

source/SH3/graphics/msbmp.cpp
Quaker762: I've done it again haha

include/SH3/graphics/msbmp.hpp
z33ky: My parser throws exceptions trying to read this sentence :S

source/SH3/graphics/msbmp.cpp
Quaker762: > data should get the same treatment.

Data is just 8-bit indexes though??

include/SH3/graphics/msbmp.hpp
Quaker762: [Doesn't is suck that English is my first language...](https://heavyeditorial.files.wordpress.com/2014/03/english-fail-cover.jpg?quality=65&strip=all&w=640)

include/SH3/graphics/msbmp.hpp
z33ky: Interesting. What do you get running this program: https://gist.github.com/z33ky/5ba9cf15d02615e62ae7927afaf1728d

Quaker762: 8 and 8....

source/SH3/graphics/msbmp.cpp
z33ky: You're uploading it as a RGB texture to OpenGL, so I thought it would be RGB color information.

include/SH3/graphics/msbmp.hpp
z33ky: Alright then, we shall just keep this for now.

include/SH3/graphics/msbmp.hpp
z33ky: Do we even need to keep the pixel data after we've uploaded it to OpenGL?

source/SH3/graphics/msbmp.cpp
Quaker762: Yeah it is sorry. I've had a long day I need some sleep haha.

include/SH3/graphics/msbmp.hpp
Quaker762: [It turns out we don't!!!!](https://www.opengl.org/discussion_boards/showthread.php/131495-glTexImage2D-use-or-copy-pixel-data)

I'll move this to a local variable, and put the `pixel` struct inside `msbmp.cpp`, though when we get some more common types going (vertex, etc), we can probably use that instead.

source/SH3/graphics/msbmp.cpp
z33ky: Yeah, but you're using `i` to write 3 color components into `data`. `data` does not contain the palette indices.

source/SH3/graphics/msbmp.cpp
z33ky: I shall try to avoid using "Farbe" as well ;)

source/SH3/graphics/msbmp.cpp
Quaker762: I just realised this, oops...

source/SH3/graphics/msbmp.cpp
Quaker762: How would I use that for this line then??

`(char*)&palette[0]`

source/SH3/graphics/msbmp.cpp
Quaker762: Hahaha classic. I can only imagine what the original source for SH3 looks like. If you open up the executable in a debugger and search for strings, you can see some realllllly Japanese style English for the error messages that makes them impossible to interpret.

source/SH3/graphics/msbmp.cpp
z33ky: `reinterpret_cast<char*>(&palette[0])`. Or `palette.data()` instead of the more C-like `&palette[0]`; the effect is the same though.

Quaker762: Ahh okay, I remember why I used the C style cast now, I can't compile using `palette.data()`

z33ky: Btw the C++ style is safer, since it doesn't allow removing `const` or `volatile` (you need `const_cast` for this, making it explicit).

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+            color = palette[index];
+            data[i].r = (color >> 8) & 0xFF; // Extract Red
+            data[i + 1].g = (color >> 16) & 0xFF; // Extract Green
+            data[i + 2].b = (color >> 24) & 0xFF; // Extract Blue
+        }
+    }
+    else if(bpp == 24)
+    {
+        file.seekg(sizeof(fheader) + sizeof(iheader) + iheader.palette_size * 4, std::ios_base::beg); // Seek to the image data
```
z33ky: Redundant seek (line 64).

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,111 @@
...
+        palette.resize(iheader.palette_size);
+        file.seekg(sizeof(fheader) + sizeof(iheader), std::ios_base::beg); // Seek to the palette
+        file.read((char*)&palette[0], iheader.palette_size);
+        file.seekg(sizeof(fheader) + sizeof(iheader) + iheader.palette_size * 4, std::ios_base::beg); // Insure we are actually in the data section
+
+        // Slightly expensive, but we now copy each pixel into the data buffer by getting it's index in the palette.
+        // Seeing as this is done once, I doubt it matters too much, especially if we do this during load time
+        // and keep the image alive for the whole of the scene duration (there are 3 bmps total in SH3, not including
+        // our default 'error' texture)
+        for(int i = 0; i < data.size(); i += 3)
```
z33ky: This is incorrect now, because `data` has become all color components.
```
for(pixel& p: data)
{
    std::uint8_t index;
    file.read(reinterpret_cast<char*>(&index), sizeof(index));
    p = palette[index];
}
```

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,111 @@
...
+    bpp = iheader.bpp;
+
+    data.resize(width * height * 3); // Some programs misformat this, so calculate it ourselves. This is usually 24-bit even for 8-bit paletted images!
+
+    // We can now differentiate if our bitmap is 8bpp or 24bpp
+    if(bpp == 8) // Paletted images are for fuckwits, I'm looking at you, Konami....
+    {
+        palette.resize(iheader.palette_size);
+        file.seekg(sizeof(fheader) + sizeof(iheader), std::ios_base::beg); // Seek to the palette
+        file.read((char*)&palette[0], iheader.palette_size);
```
z33ky, line 69: Is each element in the palette 32 bit (RGBA) or 24 bit (RGB)? Currently we expect 32 bit components...

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,111 @@
...
+            color = palette[index];
+            data[i].r = (color >> 8) & 0xFF; // Extract Red
+            data[i + 1].g = (color >> 16) & 0xFF; // Extract Green
+            data[i + 2].b = (color >> 24) & 0xFF; // Extract Blue
+        }
+    }
+    else if(bpp == 24)
+    {
+        file.seekg(sizeof(fheader) + sizeof(iheader) + iheader.palette_size * 4, std::ios_base::beg); // Seek to the image data
+        file.read((char*)&data[0], data.size()); // Read in the image data
```
z33ky: If `bpp == 24` then each pixel should be 3 bytes in size, however, `sizeof(data[0]) == 4`.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,111 @@
...
+    }
+    else
+    {
+        Log(LogLevel::WARN, "msbmp::Load( ): Bad pixel depth %d!", bpp);
+        return res;
+    }
+
+    glGenTextures(1, &tex); // Bind texture for use
+    glBindTexture(GL_TEXTURE_2D, tex);
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, width, height, 0, GL_RGB, GL_UNSIGNED_BYTE, &data[0]);
```
z33ky: `data` contains RGBA pixels, so this should be `GL_RGBA` (or `pixel` should be changed to only have RGB components).

z33ky (changes requested):

source/SH3/graphics/msbmp.cpp
Quaker762: Each colour in the palette is 4 bytes, but has 3 components.... IIRC BMP doesn't support having an alpha channel. Open up error.bmp in a hex editor and you'll see. I'll rename a to unused asap (technically alpha because it's hard locked to 255, full colour)

EDIT: I accidentally added a in the pixel struct. I'll remove it asap.

source/SH3/graphics/msbmp.cpp
Quaker762: For some reason if I omit this its starts reading indexes from the palette... I'll fiddle with this later.

source/SH3/graphics/msbmp.cpp
Quaker762: Crap, I'll remove the alpha ASAP

source/SH3/graphics/msbmp.cpp
Quaker762: Shit I didn't realise this. I'll change it ASAP

source/SH3/graphics/msbmp.cpp
Quaker762: The colour palette (which is still contained in a 24bpp bmp IIRC) is comprised of 4 byte colours though 3 are used. I'll double check this with a 24bpp bmp tomorrow.

source/SH3/graphics/msbmp.cpp
z33ky: Ah nah, my error. Line 64 is in the `bpp == 8` branch.

Quaker762:
Well, it turns out the colour palette is ARGB stored in little endian..

Also, the image data is actually UPSIDE DOWN, as the origin is from the bottom left.. I'll work on something to flip it.

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    data.resize(width * height * 3); // Some programs misformat this, so calculate it ourselves. This is usually 24-bit even for 8-bit paletted images!
+
+    // We can now differentiate if our bitmap is 8bpp or 24bpp
+    if(bpp == 8) // Paletted images are for fuckwits, I'm looking at you, Konami....
+    {
+        palette.resize(iheader.palette_size);
+        file.seekg(sizeof(fheader) + sizeof(iheader), std::ios_base::beg); // Seek to the palette
+        file.read((char*)&palette[0], iheader.palette_size);
+        file.seekg(fheader.pix_offset, std::ios_base::beg); // Insure we are actually in the data section
```
z33ky: Ensure*

source/SH3/graphics/msbmp.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+msbmp::load_result msbmp::Load(const std::string& path)
+{
+    std::ifstream   file;
+    bmp_file_header fheader;    // File information header
+    bmp_info_header iheader;
+    load_result     res;
+    std::uint16_t   bpp;        // Bitmap bit depth
+
+    std::vector<std::uint32_t> palette;     // Vector containing colour data for this image
```
z33ky:
```
struct palette_color
{
    std:;uint8_t b, g, r, a;
};
```

z33ky (changes requested):

source/SH3/graphics/msbmp.cpp
Quaker762: I'll change it, but you can use both (especially if you want [American English](https://en.oxforddictionaries.com/usage/ensure-or-insure)) ;) Ensure is more commonly used though, I'm not sure how I made that error.... The only thing worse than the code in this PR is my spelling AHAHA

source/SH3/graphics/msbmp.cpp
Quaker762: Is struct reading faster than bitmasking?? (not that it really matters with the processor speeds of today ahaha).

Quaker762:
Also, now that we have a few formats ready to go, we should probably focus on testing drawing/getting some front end engine stuff going. After that, there's two image formats left, `tbn2` and `pic` (which I'm working on atm, it's AKA SoftImage PIC). I think `tbn2` is a regular texture for effects, but PIC I have no idea what the usage is. After that models, shaders and skeletal animation is the last of the graphics submodules (oh, and lighting and shadowing ;] ).

source/SH3/graphics/msbmp.cpp
z33ky: It might use byte-sized moves instead of bit twiddling, unless the optimizer decides that the latter is more efficient on the platform (keeping [this](http://wiki.c2.com/?SufficientlySmartCompiler) in mind).
Using a struct results in less magic constants though.

z33ky (approved):

PR #82, opened 2017-02-10
Added second version of sh3_glprogram::Load()

Quaker762:
glprograms can now take attributes and bind them, so we can actually now send stuff to the vertex shader (like vertex normals etc). Requires a second constructor and Load( ) implementation.

### DIFF
--- a/include/SH3/system/glprogram.hpp
+++ b/include/SH3/system/glprogram.hpp
@@ -13,10 +13,15 @@
 #ifndef GLPROGRAM_HPP_INCLUDED
 #define GLPROGRAM_HPP_INCLUDED
 
+#include "SH3/error.hpp"
+
 #include <GL/glew.h>
 #include <GL/gl.h>
 
 #include <string>
+#include <vector>
+
+
 
 namespace sh3_graphics
 {
@@ -24,18 +29,38 @@ namespace sh3_graphics
     struct sh3_glprogram final
     {
     public:
+        enum class load_result
+        {
+            SUCCESS,
+            COMPILATION_FAILED,
+            LINK_FAILED,
+            INVALID_VALUE,
+            INVALID_OPERATION,
+        };
 
-        sh3_glprogram(const std::string& name) : programName(name){Load(name);};
+        class load_error final : public error<load_result>
+        {
+        public:
+            void set_error(load_result result);
+
+            std::string message() const;
+
+        private:
+        };
+
+        sh3_glprogram(const std::string& name, load_error& err, const std::vector<std::string>& attribs = {}) : programName(name){Load(name, err, attribs);};
         ~sh3_glprogram(){Unbind(); glDeleteProgram(programID);}
 
         /**
-         *  Load a shader from a file on disk, compile and then link it.
+         *  Load a shader from a file on disk, compile and then link it, binding any attributes in the process.
          *
          *  @param shader Name of the shader we want to load (path is hard-coded to /data/shaders). Assumes *.vert and *.frag have the same name.
+         *  @param attribs Vector of attributes we bind before we link the program.
+         *  @param err Error set by this operation.
          *
-         *  @return GLint true or false depending on whether this shader was loaded correctly.
+         *  @return @ref load_error The error set by this operation (specified in the @ref load_result enum)
          */
-        GLint Load(const std::string& shader);
+        void Load(const std::string& shader, load_error& err, const std::vector<std::string>  &attribs = {});
 
         /**
          *  Bind this shader program for use.
--- a/source/SH3/system/glprogram.cpp
+++ b/source/SH3/system/glprogram.cpp
@@ -18,9 +18,39 @@ static const GLuint BAD_SHADER = 0;
 
 using namespace sh3_graphics;
 
-GLint sh3_glprogram::Load(const std::string& name)
+void sh3_glprogram::load_error::set_error(load_result res)
 {
-    GLint status = 0;
+    result = res;
+}
+
+std::string sh3_glprogram::load_error::message() const
+{
+    std::string error;
+    switch(result)
+    {
+    case load_result::SUCCESS:
+        error = "SUCCESS";
+        break;
+    case load_result::COMPILATION_FAILED:
+        error = "COMPILATION_FAILED";
+        break;
+    case load_result::LINK_FAILED:
+        error = "LINK_FAILED";
+        break;
+    case load_result::INVALID_VALUE:
+        error = "GL_INVALID_VALUE";
+        break;
+    case load_result::INVALID_OPERATION:
+        error = "GL_INVALID_OPERATION";
+        break;
+    }
+
+    return error;
+}
+
+void sh3_glprogram::Load(const std::string& shader, load_error& err, const std::vector<std::string>& attribs)
+{
+    GLenum status = 0;
 
     programID = glCreateProgram();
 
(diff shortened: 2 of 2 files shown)


source/SH3/system/glprogram.cpp
```diff
@@ -55,6 +55,50 @@ GLint sh3_glprogram::Load(const std::string& name)
     glDeleteShader(fragShader);
 }
 
+GLint sh3_glprogram::Load(const std::string& name, std::vector<std::string> const &attribs)
```
z33ky: This is a huge code duplication with `Load(std::string)`. We can just default `attribs` to an empty vector (`std::vector<std::string> const &attribs = {}`) and it'll do the same as the other `Load` method.

include/SH3/system/glprogram.hpp
```diff
@@ -26,6 +27,7 @@ namespace sh3_graphics
     public:
 
         sh3_glprogram(const std::string& name) : programName(name){Load(name);};
+        sh3_glprogram(const std::string& name, std::vector<std::string> const &attribs) : programName(name){Load(name, attribs);};
```
z33ky: If we unify the `Load` methods, we can also unify the constructors the same way.

source/SH3/system/glprogram.cpp
```diff
@@ -55,6 +55,50 @@ GLint sh3_glprogram::Load(const std::string& name)
...
+    vertShader = (GLuint)Compile(GL_VERTEX_SHADER);
+    fragShader = (GLuint)Compile(GL_FRAGMENT_SHADER);
+
+    glAttachShader(programID, vertShader);
+    glAttachShader(programID, fragShader);
+
+    // Bind any attributes before we link the program
+    for(int i = 0; i < attribs.size(); i++)
+    {
+        glBindAttribLocation(programID, i, reinterpret_cast<const GLchar*>(attribs[i].c_str()));
```
z33ky, line 67: Should we check for errors here?

z33ky (changes requested):

source/SH3/system/glprogram.cpp
Quaker762: Done. I forgot you can do this, I'm still getting used to some more modern c++ stuff sorry.

source/SH3/system/glprogram.cpp
Quaker762: What errors are you referring to?? As far as I know there aren't any that can occur. It's usually just link and compile that can crap the bed.

include/SH3/system/glprogram.hpp
Quaker762: So should I default them to empty using `std::vector<std::string> const &attribs = {}` in the constructor and method, or is just the constructor fine?

source/SH3/system/glprogram.cpp
z33ky: > `GL_INVALID_VALUE` is generated if *index* is greater than or equal to `GL_MAX_VERTEX_ATTRIBS`.

I was expecting it to also set an error if the attribute name cannot be found, but the spec doesn't mention that case.

include/SH3/system/glprogram.hpp
z33ky: Hm. Is it even useful to have `Load` as a separate method?

I also just noticed that the `const` is written in different locations for both parameters. It'll have the same effect, but I'd prefer a consistent style.

source/SH3/system/glprogram.cpp
```diff
@@ -55,6 +55,50 @@ GLint sh3_glprogram::Load(const std::string& name)
...
+
+    // Generate our vertex and fragment shader
+    vertShader = (GLuint)Compile(GL_VERTEX_SHADER);
+    fragShader = (GLuint)Compile(GL_FRAGMENT_SHADER);
+
+    glAttachShader(programID, vertShader);
+    glAttachShader(programID, fragShader);
+
+    // Bind any attributes before we link the program
+    for(int i = 0; i < attribs.size(); i++)
```
z33ky: We should use a `std::size_t` for counting.

z33ky (changes requested):
The spec for `glBindAttribLocation` mentions aliasing, i.e. binding the same index to multiple attributes.
If this is something we want to have, we should use a `std::map<std::string /*attribute*/, GLuint /*index*/>` instead of the `std::vector`.

include/SH3/system/glprogram.hpp
Quaker762: Perhaps not, I was doing it to keep it consistent with other modules and to add a way to attempt to "reload" if something fails. I'll move the code for `Load` into the constructor and move the constructor into `glprogram.cpp` 

EDIT: I just realised I was using a return code for `Load`. Should I move the code into the constructor or keep it as it is?

source/SH3/system/glprogram.cpp
Quaker762: You can have any attribute name you want I believe. I think attributes are sort of like executable symbols like `.rodata`, `.text` and `.data` (though I'm not 100% certain on this), which is why they need to be specified pre link. This is the only way the CPU can actually interface with the GPU to send it data... I still don't have a clear picture though.. An  @error might get set trying to call `glGetAttribLocation` with an invalid name though.

I'll add some error checking for all the errors it can throw, I'm sure it probably will across different graphics cards, God knows the differences nVidia and AMD have across their drivers.... @

Quaker762:
> The spec for glBindAttribLocation mentions aliasing, i.e. binding the same index to multiple attributes.

I'm not sure why we'd want to do this?? Realistically, all we're going to send to the shader is normals, uv co-ords etc (perhaps bone weights for skeletal animation). The only reason I can see this being a problem is with nVidia reserving some names with indexes (such as gl_Normal, with an explanation [here](http://www.stevestreeting.com/2008/04/26/working-around-glsl-attribute-aliasing-problems-on-os-x/) ). 

Apart from that I actually have no clue why someone would want to bind two attributes to the same location.

include/SH3/system/glprogram.hpp
z33ky: Return the error in an output-parameter, like it's done for `sh3_arc_file::Read*`.

source/SH3/system/glprogram.cpp
z33ky: What I meant is that you might accidentally specify an attribute name that does not exist in the shader, for example because of misspelling, or renaming an attribute in the shader source and forgetting recompile the game binary.

Quaker762: Oh, oops sorry. I didn't even think of that happening ahaha. I'll do a few check and logs for the errors then.

z33ky:
> I'm not sure why we'd want to do this??

I can't think of anything on the spot. I just found it curious that it's allowed and thought maybe it's something people commonly do. We can always add it later, when we do find a use-case.

> The only reason I can see this being a problem is with nVidia reserving some names with indexes (such as gl_Normal, with an explanation [here](http://www.stevestreeting.com/2008/04/26/working-around-glsl-attribute-aliasing-problems-on-os-x/) ).

Here's to hoping Vulkan won't be an archaic mess of backwards-compatibility in 10 years.

Quaker762:
Sounds good. It's probably something they put it to fix a hack that fixed a hack that fixed a hack from 1996 lol.

> Here's to hoping Vulkan won't be an archaic mess of backwards-compatibility in 10 years.

One can only hope hahaha.

include/SH3/system/glprogram.hpp
```diff
@@ -13,29 +13,54 @@
...
+            INVALID_VALUE,
+            INVALID_OPERATION,
+        };
+
+        class load_error final : public error<load_result>
+        {
+        public:
+            void set_error(load_result result);
+
+            std::string message() const;
```
z33ky, line 46: Implementation is missing.

z33ky (changes requested):

include/SH3/system/glprogram.hpp
Quaker762: Fuckin hell hahaha...

z33ky (approved):

PR #83, opened 2017-02-11
Created a quad class

Quaker762:
Created a class that enables us to actually draw a quad to the screen
(and a small GL3.3 test in main.cpp). Currently doesn't pass any data to
a shader, so it's just a black quad (no colour data).

### DIFF
--- a/include/SH3/graphics/quad.hpp
+++ b/include/SH3/graphics/quad.hpp
@@ -1,0 +1,47 @@
+/** @file
+ *
+ *  Class to render a fullscreen 2D quad to the screen.
+ *
+ *  @copyright 2017 Palm Studios
+ *
+ *  @date 11-2-2017
+ *
+ *  @note To render a quad, we need to split the screen into two right-angled triangles, meaning we need 6 vertices for one quad.
+ *  @note We're not using Index Buffering here because a quad only has 6 verts... Kind of a waste of time,
+ *
+ *  @author Jesse Buhagiar
+ */
+#ifndef QUAD_HPP_INCLUDED
+#define QUAD_HPP_INCLUDED
+
+#include "GL/glew.h"
+#include "GL/gl.h"
+#include "glm/glm.hpp"
+#include "SH3/types/vertex.hpp"
+
+#include <array>
+
+namespace sh3_graphics
+{
+    struct quad final
+    {
+    public:
+        /**
+         *  Class constructor. Takes
+         */
+        quad(const std::array<vertex3f, 6>& verts);
+        ~quad(){};
+
+        /**
+         *  Draw this quad to the screen
+         */
+        void Draw();
+
+    private:
+        GLuint vao;     /**< VAO where we will store the format of the vertex data and buffer objects */
+        GLuint vbuff;   /**< The VBO we fill with our vertex data */
+    };
+}
+
+
+#endif // QUAD_HPP_INCLUDED
--- a/include/SH3/types/vertex.hpp
+++ b/include/SH3/types/vertex.hpp
@@ -12,32 +12,37 @@
 
 #include <GL/gl.h>
 
+#pragma pack(push, 1)
+
+/** @defgroup graphics-types Graphics Types
+ *  @{
+*/
+
+/** UV Texture Co-Ordinate  structure*/
+struct texcoord
+{
+    GLfloat u, v;
+};
+
 /** Basic vertex type */
-typedef struct
+struct vertex3f
 {
     GLfloat x, y, z;      /**< Position */
-} vertex3f_t;
+};
 
-/** Textured Vertex with normals */
-typedef struct
+/** Model Vertex (for static models) */
+struct sh3_model_vertex
 {
-    Glfloat x, y, z;
-    Glfloat u, v;         /**< Texture UV co-ords */
-    GLfloat nx, ny, nz;   /**< Vertex normals (do we need 3??) */
-} vertex3f_tex_t;
+    vertex3f vertex;    /**< Vertex */
+    vertex3f normal;    /**< Vertex Normal */
+    texcoord uvcoord;   /**< Texture UV Co-Ordinate for this vertex */
+};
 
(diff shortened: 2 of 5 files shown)


Quaker762:
@z33ky I think I _finally_ understand how to draw stuff in OpenGL haha. I'll work on the model class soon, should be a cinch porting Mike's code over to some nicer C++ (if you can consider my code that AHAHA)

include/SH3/graphics/quad.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace sh3_graphics
+{
+    struct quad final
+    {
+    public:
+        /**
+         *  Class constructor. Takes
+         */
+        quad(const std::vector<glm::vec3>& verts);
```
Quaker762: I think we should change this to our type in `vertex.hpp` which would be `vertex3f_t`

z33ky: We could also alias `vertex3f_t` as `glm::vec3`.

include/SH3/graphics/quad.hpp
```diff
@@ -0,0 +1,45 @@
...
+        ~quad(){};
+
+        /**
+         *  Draw this quad to the screen
+         */
+        void Draw();
+
+    private:
+        GLuint vao; /**< VAO where we will store the four co-ordinates for our 2D quad in slot 1 ("position") */
+        GLuint vbo; /**< The buffer we will */
```
z33ky: "we will" ..,

include/SH3/graphics/quad.hpp
z33ky: This could be a `std:array<glm::vec3, 4>`.

include/SH3/graphics/quad.hpp
Quaker762: I think I went off implementing before I finished my documentation lol.

include/SH3/graphics/quad.hpp
Quaker762: I think it's superfluous unless we're doing some math. All we need is 3 points, the type `vertex3f` makes more sense imho.

Quaker762: Okay I'll work on indexing then, otherwise we'd need 6 vertexes (2 right-angled triangles)

Quaker762:
I haven't bothered with indexing for the quad (aka I couldn't get it to work haha), it's a waste of time trying to get it going for a saving of a few bytes of VRAM..

Quaker762:
BTW shader programs are working rather nicely :)

Shaders!

include/SH3/graphics/quad.hpp
z33ky: Hm, I guess unless we want to implement our own physics we indeed won't need many vector operations.

z33ky: I think I see the issue. We'd need to specify which direction the quad faces.
6 will work for now, until we figure out how we want to use this class.

 @Edit:
We could also take `std::array<glm::vec3, 3> center, std::array<glm::vec3, 3> normal, glm::vec3::length_t width, glm::vec3::length_t height`, but yeah, let's first figure out how we'll use this.

include/SH3/graphics/quad.hpp
```diff
@@ -0,0 +1,47 @@
...
+        quad(const std::array<vertex3f, 6>& verts);
+        ~quad(){};
+
+        /**
+         *  Draw this quad to the screen
+         */
+        void Draw();
+
+    private:
+        GLuint vao;     /**< VAO where we will store the four co-ordinates for our 2D quad in slot 1 ("position") */
```
z33ky: Doesn't the VAO store the *format*, not the actual coordinates?

include/SH3/graphics/quad.hpp
Quaker762: That would actually be a good feature to add instead of death animations, ragdoll physics haha.

But yeah, I agree. I'm starting to think it's pointless even having `glm` to some extent, as we can probably roll our own matrices and remove a dependency entirely.

Quaker762: > until we figure out how we want to use this class.

I was thinking along the lines of menu backgrounds and decals, though I'm not sure how Mike originally has this done (i.e the save icon on a wall)

> We could also take std::array<glm::vec3, 3> center

Wouldn't it be easier to just take 4 vertexes and have the indices pre-calculated?? They shouldn't really need to change for a quad.

include/SH3/graphics/quad.hpp
Quaker762: Very true, I was mistaken in writing this. I'll change it ASAP.

Quaker762:
As a side note (sort of unrelated), it seems the indices for models are baked into the vertex data headers in all `.mdl` files.

include/SH3/graphics/quad.hpp
z33ky: > decals

That would only work on flat surfaces.

> Wouldn't it be easier to just take 4 vertexes and have the indices pre-calculated??

Depends. If we have places where we know the position the quad should be in it'll be better to have the code for calculating the coordinates of the vertices in one place instead of having to replicate it on the caller site.

Quaker762: Ahh I see. I'll have to look at Mikes scene loader (I believe I sent it to you, it'll be in the file SH3_Loader.hpp/cpp), it's a bit of a mess so it's slightly hard to read.

For the most part I think it'll just be triangles and triangle strips, I'm not too sure of the efficieny of using a quad.

z33ky: > For the most part I think it'll just be triangles and triangle strips, I'm not too sure of the efficieny of using a quad.

Our quad uses triangles internally, so this is mute.

Quaker762: Oh sorry, I thought you were referring to drawing more complex geometry using quads.

PR #87, opened 2017-02-21
Updated texture class (again)

Quaker762:
Everything works as it should now (I hope). 8-bit textures have been
tested and are confirmed to be working (for the most part), however, 16,
24, and 32 bit textures have not, but they should work just fine.

I think the `normal` and `transform` arrays need some tweaking. The output is slightly corrupt/offset a few pixels to the right, and I'm sure that's the culprit. That or the palette is encrypted as well (I hope not...)

### DIFF
--- a/source/SH3/graphics/texture.cpp
+++ b/source/SH3/graphics/texture.cpp
@@ -9,13 +9,110 @@
  *  @author Jesse Buhagiar
  */
 #include <SH3/graphics/texture.hpp>
+#include <SH3/system/assert.hpp>
 #include <SH3/system/log.hpp>
 #include <SH3/arc/vfile.hpp>
 
+#include <algorithm>
 #include <cassert>
-
+#include <fstream>
+#include <functional>
+#include <limits>
 
 using namespace sh3_graphics;
+
+#pragma pack(push, 1)
+struct rgba
+{
+    std::uint8_t r, g, b, a;
+};
+
+struct rgba16
+{
+    std::uint8_t r : 5;
+    std::uint8_t g : 5;
+    std::uint8_t b : 5;
+    std::uint8_t a : 1;
+};
+
+struct rgb
+{
+    std::uint8_t r, g, b;
+};
+
+struct bgra
+{
+    std::uint8_t b, g, r, a;
+};
+
+struct bgr
+{
+    std::uint8_t b, g, r;
+};
+
+struct tga_header
+{
+    static constexpr std::uint8_t TYPE_RGB24 = 2;
+    static constexpr std::uint8_t FLAGS_FLIP = 0x20;
+
+    std::uint8_t id_size = 0;               /**< Size of the ID field that follows this header (usually 0) */
+    std::uint8_t colormap = 0;              /**< Is this image paletted (THIS MUST ALWAYS BE 0 FOR US!) */
+    std::uint8_t type = TYPE_RGB24;         /**< Type of image (ALWAYS 2 FOR RGB24!) */
+    std::uint8_t unused[5] = {};
+
+    std::uint16_t x_origin = 0;             /**< Co-ordinate for first X value */
+    std::uint16_t y_origin = 0;             /**< Co-ordinate for first Y value */
+    std::uint16_t width = 0;
+    std::uint16_t height = 0;
+    std::uint8_t bpp = 24;                  /**< Bits per pixel */
+    std::uint8_t flags = FLAGS_FLIP;
+};
+
+#pragma pack(pop)
+
+namespace
+{
+/**
+ *  Dump a texture to a TARGA/TGA file
+ *
+ *  @param width - The Width of this texture.
+ *  @param height - The Height of this texture.
+ *  @param data - Pixel data
+ *  @param bpp - Bitness of data (not the actual texture).
+ */
+void DumpRGB2Bitmap(std::uint32_t width, std::uint32_t height, std::vector<std::uint8_t>& data, std::uint8_t bpp)
+{
+    tga_header header;
+    std::ofstream file("output.tga", std::ios::binary); // Open the stream for binary output
+
+    ASSERT(width <= std::numeric_limits<decltype(header.width)>::max());
+    ASSERT(height <= std::numeric_limits<decltype(header.height)>::max());
+
+    header.width        = static_cast<std::uint16_t>(width);
+    header.height       = static_cast<std::uint16_t>(height);
+
+    switch(bpp)
+    {
+    case 16:
+        header.bpp = 16;
+        break;
+    case 24:
+        header.bpp = 24;
+        break;
+    case 32:
+        header.bpp = 32;
+        break;
+    default:
+        Log(LogLevel::WARN, "DumpRGB2Bitmap( ): Warning: Invalid bpp passed to function. Not writing image data.");
+        return;
+        break; // Do nothing as we have a default value of 32 (RGB).
+    }
+
+    file.write(reinterpret_cast<char*>(&header), sizeof(header));
+    file.write(reinterpret_cast<char*>(&data[0]), data.size());
+}
+}
+
 
 //TODO: Scale the texture and then
 void sh3_texture::Load(sh3_arc& mft, const std::string& filename)
@@ -23,43 +120,279 @@ void sh3_texture::Load(sh3_arc& mft, const std::string& filename)
     sh3_texture_header          header;
     sh3_arc_vfile               file(mft, filename);
     sh3_arc_vfile::read_error   e;
-    std::vector<std::uint8_t>   data;     // Pixel data of this texture (with the header stripped)
+    std::vector<std::uint8_t>   data;       // Pixel data of this texture (with the header stripped)
 
-    std::size_t ret = file.ReadData(&header, sizeof(header), e); // Read in the header
+    std::streamsize             offset = 0;
+
+    std::size_t ret = file.ReadData(&header, sizeof(header), e);
 
     if(ret != sizeof(header))
-        die("sh3_texture::Load( ): Header only read %d bytes of data! It should have been %d", ret, sizeof(header)); // TODO: Fallback on dummy texture (like Source's purple and black checkboard)
+        die("sh3_texture::Load( ): ReadData( ) != sizeof(header)!");
 
-    if(header.bpp == 8)
-        Log(LogLevel::WARN, "WARNING: 8-bit textures are currently unsupported! They WILL look like nothing but pixel garbage.");
-
-    Log(LogLevel::INFO, "Texture info:\n\ttexWidth:%d\n\ttexHeight:%d\n\ttexSize:%d\n\tbpp:%d", header.texWidth, header.texHeight, header.texSize, header.bpp);
-
-    file.Seek(sizeof(header) + header.dataOffset, std::ios_base::beg);
-
-    data.resize(header.texSize);
-
-    if(header.texSize != header.texWidth * header.texHeight * header.bpp)
+    // Check for the pesky 64-byte A7A7A7A7 header that sometimes precedes our texture header
+    if(header.batchHeaderMarker == 0x00000000 && header.batchSize == 0xA7A7A7A7) // AHA!
     {
-        Log(LogLevel::ERROR, "sh3_texture::Load( ): texWidth * texHeight * bpp != header.texSize!");
-        return;
+        offset = 0x40; // Skip the unknown header if it exists
+
+        file.Seek(offset, std::ios_base::beg);
+        file.ReadData(&header, sizeof(header), e);
+
+        if(ret != sizeof(header))
+            die("sh3_texture::Load( ): ReadData( ) != sizeof(header)!");
     }
 
-    ret = file.ReadData(&data[0], header.texSize, e);
-    if(ret != header.texSize)
-        Log(LogLevel::WARN, "sh3_texture::Load( ): Only read %d out of %d bytes!", ret, header.texSize);
+    if(header.texSize == header.texWidth * header.texHeight* 4)
+    {
+        header.bpp = 32; // Thanks KONAMI!
+    }
 
-    // Now actually create a logical texture with OpenGL on the gpu
-    glGenTextures(1, &tex);
+    // Now that we're done that, we can check perform some sanity checks on our texture!
+    if(header.texSize != header.texWidth * header.texHeight * header.bpp / 8)
+    {
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Warning, texSize != width * height * (bpp / 8)!");
+        return; // TODO: Bind a color shader here
+    }
 
-    glBindTexture(GL_TEXTURE_2D, tex);
-    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, header.texWidth, header.texHeight, 0, GL_BGRA, GL_UNSIGNED_BYTE, &data[0]);
-    glBindTexture(GL_TEXTURE_2D, 0); // Unbind this texture
+    data.resize(header.texSize); // Early data resize (if it's an 8bpp texture, it will be resized anyway)
+
+    if(header.bpp == PixelFormat::PALETTE)
+    {
+        palette_info         pal_header;
+        std::vector<bgra>    palette;    // Palette Data (I think this is BGRA)
+
+        // First, we need to seek to the palette and read it in.
+        file.Seek(offset + header.batchHeaderSize + header.texFileSize, std::ios_base::beg);
+        file.ReadData(&pal_header, sizeof(pal_header), e);
+
+        // Palette information is stored in blocks (usually of size 64-bytes). We also know how large the
+        // palette is (in bytes, including padding between blocks). From this, we can deduce (with a bit of math)
+        // that the whole palette occupies ~ paletteSize/entrySize bytes of space, contains a total of
+        // entrySize/bypp colors per block, which therefore means we have a total of nBlocks * col_per_block colors,
+        // which equates to about 256-colors in total (which seems accurate for an 8-bit texture).
+
+        const std::size_t nBlocks = (pal_header.paletteSize / pal_header.entrySize) / pal_header.bytes_per_pixel;
+        const std::size_t colorsPerBlock = pal_header.entrySize / pal_header.bytes_per_pixel;
+
+        palette.resize(colorsPerBlock * nBlocks);
+
+        for(std::size_t block = 0; block < nBlocks; ++block)
+        {
+            std::size_t read = file.ReadData(static_cast<void*>(&palette[block * colorsPerBlock]), pal_header.entrySize, e);
+
+            if(read != pal_header.entrySize)
+            {
+                Log(LogLevel::WARN, "sh3_texture::Load( ): Warning: Number of bytes read in palette block != pal_header.entrySize! (Expected %d, got %d)", read, pal_header.entrySize);
+                break;
+            }
+
+            // Each block is 256 bytes.
+            // Seek to next block.
+            file.Seek(256 - pal_header.entrySize, std::ios_base::cur);
+        }
+
+        // we need to swap some colors every 32 pixels, starting from the 8th
+        static constexpr std::size_t swapDistance = 32; // distance between swaps
+        static constexpr std::size_t swapSize = 8; // amount of colors to swap
+
+        if(palette.size() > 8)
+        {
+            for(auto iter = next(begin(palette), 8); distance(iter, end(palette)) > swapDistance; advance(iter, swapDistance))
+            {
+                // swap 8 colors
+                const auto swapBlock = next(iter, swapSize);
+                if(distance(swapBlock, end(palette)) < swapSize)
+                {
+                    Log(LogLevel::WARN, "Palette doesn't have enough colors left for swapping.");
+                    break;
+                }
+                std::swap_ranges(iter, swapBlock, swapBlock);
+            }
+        }
+
+
+        // Now that we've completely loaded the palette in its entirety, we can get the 8-bit index value
+        // from the data section of the file and get it's color in the palette!
+
+        //===---THIS IS A CLUSTER FUCK FOR NOW UNTIL WE UNDERSTAND HOW IN THE NAME OF CHRIST THIS WORKS---===//
+        std::vector<std::uint8_t> iBuffer;  // Our index buffer that we put transformed indecies into
+
+        data.resize(header.texWidth * header.texHeight * 3); // We strip the Alpha channel from the BGRA pixel beacuse it is hard locked to 0x80 (not 0xFF!!)
+        iBuffer.resize(header.texSize);
+
+        file.Seek(offset + (header.texFileSize - header.texSize), std::ios_base::beg); // Seek to the beginning of data
+
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            if(header.texWidth % 16 != 0)
+            {
+                Log(LogLevel::WARN, "sh3_texture::Load( ): Warning: texWidth not divisible by 16!");
+                header.texWidth -= header.texWidth % 16;
+            }
+
+            if(header.texHeight % 4 != 0)
+            {
+                Log(LogLevel::WARN, "sh3_texture::Load ( ): Warning: texHeight not divisible by 4!");
+                header.texHeight -= header.texHeight % 4;
+            }
+
+            std::uint32_t x = 0;
+            std::uint32_t y = 0;
+            bool offsetFlipper = false;
+
+            // We need to step through each index and unencrypt it
+            // FIXME: distortion on 16-pixel wide block on the left
+            while(true)
+            {
+                for(unsigned i = 0; i < 32; ++i)
+                {
+                    std::uint8_t index;
+                    file.ReadData(&index, sizeof(index), e);
+
+                    std::uint8_t xoffset = ((i << 2) & 0xf) + ((i >> 2) & 0xf);
+                    if(i > 16 && i % 2) // aka (i & 17) == 17
+                    {
+                        xoffset ^= 8;
+                        xoffset &= 0xf;
+                    }
+                    if(offsetFlipper)
+                    {
+                        xoffset ^= 4;
+                    }
+
+                    const auto tempx = x + xoffset; // - 16;
+                    // every other pixel is for (y + 2)
+                    const auto tempy = y + ((i % 2) ? 2 : 0);
+
+                    iBuffer[(header.texWidth * tempy) + tempx % header.texWidth] = index;
+                }
+
+                x += 16;
+                if(x < header.texWidth)
+                {
+                    continue;
+                }
+
+                x = 0;
+
+                ++y;
+                if(y % 2 == 0)
+                {
+                    // each iteration we read two lines: offset 0 and 2
+                    // so after two iterations, we have read the lines with offset 0 and 2, 1 and 3
+                    // so now we need to skip two additional lines, since we read 2 and 3 already
+                    y += 2;
+
+                    if(y >= header.texHeight)
+                    {
+                        Log(LogLevel::WARN, "sh3_texture::Load( ): Warning: y <= header.texHeight!");
+                        break;
+                    }
+
+                    offsetFlipper = !offsetFlipper;
+                }
+
+                if(y == header.texHeight)
+                {
+                    break;
+                }
+            }
+
+            // Christ help me...
+            for(std::size_t pindex = 0, doff = 0; pindex < iBuffer.size() && doff < data.size(); pindex++, doff += 3)
+            {
+                bgra pixel = palette[iBuffer[pindex]];
+
+                data[doff + 0]  = pixel.r;
+                data[doff + 1]  = pixel.g;
+                data[doff + 2]  = pixel.b;
+            }
+        }
+        else // If the distortion flag isn't set, just read the pixel data in from the palette.
+        {
+            for(std::size_t i = 0; i < header.texWidth * header.texHeight * 3; i += 3)
+            {
+                std::uint8_t index;
+                file.ReadData(&index, sizeof(index), e);
+                bgra pixel = palette[index];
+
+                data[i + 0]   = pixel.r;
+                data[i + 1]   = pixel.g;
+                data[i + 2]   = pixel.b;
+            }
+        }
+
+        DumpRGB2Bitmap(header.texWidth, header.texHeight, data, 24);
+    }
+    else if(header.bpp == PixelFormat::RGBA)
+    {
+        file.ReadData(&data[0], header.texSize, e);
+        DumpRGB2Bitmap(header.texWidth, header.texHeight, data, 32);
+    }
+    else if(header.bpp == PixelFormat::BGR)
+    {
+        file.ReadData(&data[0], header.texSize, e);
+        DumpRGB2Bitmap(header.texWidth, header.texHeight, data, 24); // Output will be reversed!
+    }
+    else if(header.bpp == PixelFormat::RGBA16)
+    {
+        //TODO: Some kind of fucked up shit here. I think this is R5G5B5A1 or something like that..
+        file.ReadData(&data[0], header.texSize, e);
+        DumpRGB2Bitmap(header.texWidth, header.texHeight, data, 16);
+    }
+    else
+    {
+        die("sh3_texture::Load( ): Unknown Pixel Format, %d", header.bpp);
+    }
+
+
+    glGenTextures(1, &tex);             // Create a texture
+    glBindTexture(GL_TEXTURE_2D, tex);  // Bind it for use
+
+    GLenum srcFormat;
+    GLint dstFormat;
+    GLenum type;
+
+    // Create the texture according to its pixel format!
+    switch(header.bpp)
+    {
+        case PixelFormat::RGBA:     // Regular 32-bit RGBA
+            srcFormat = GL_RGBA;
+            dstFormat = GL_RGBA;
+            type = GL_UNSIGNED_BYTE;
+            break;
+        case PixelFormat::BGR:      // 24-bit BGR
+            srcFormat = GL_BGR;
+            dstFormat = GL_RGB;
+            type = GL_UNSIGNED_BYTE;
+            break;
+        case PixelFormat::RGBA16:   // 16-bit RGBA. OpenGL supports this (I think)
+            srcFormat = GL_RGBA;
+            dstFormat = GL_RGBA;
+            type = GL_UNSIGNED_SHORT_5_5_5_1;
+            break;
+        case PixelFormat::PALETTE:
+            srcFormat = GL_RGB;
+            dstFormat = GL_RGB;
+            type = GL_UNSIGNED_SHORT_5_5_5_1;
+            break;
+        default:
+            die("sh3_texture::Load( ): Invalid pixel format: %d", header.bpp);
+            break;
+    }
+
+    glTexImage2D(GL_TEXTURE_2D, 0, srcFormat, header.texWidth, header.texHeight, 0, dstFormat, type, &data[0]);
+
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_REPEAT);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_REPEAT);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR); // Use linear interpolation for the texture
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+
+    glBindTexture(GL_TEXTURE_2D, tex); // Un-bind this texture.
 }
 
 void sh3_texture::Bind(GLenum textureUnit)
 {
-    assert(textureUnit >= GL_TEXTURE0 && textureUnit <= GL_TEXTURE31);
+    ASSERT(textureUnit >= GL_TEXTURE0 && textureUnit <= GL_TEXTURE31);
 
     glActiveTexture(textureUnit);
     glBindTexture(GL_TEXTURE_2D, tex);
--- a/include/SH3/graphics/texture.hpp
+++ b/include/SH3/graphics/texture.hpp
@@ -2,13 +2,14 @@
  *
  *  Functions and structures to load a texture from an arc section
  *
- *  @copyright 2016  Palm Studios and Mike M
+ *  @copyright 2016  Palm Studios, Mike Meinhardt and de_lof
  *
  *  @note   I'd like to thank Mike for all this, he put a lot of time into reverse engineering
-            all of the file types, including all of the textures, meshes and even the motion capture
-            skeletal animation that Konami captured. Thanks mate!
+ *          all of the file types, including all of the textures, meshes and even the motion capture
+ *          skeletal animation that Konami captured. Thanks mate!
  *
- *  @note 8-bit textures are currently unsupported, as they are palletted. See <a href=https://www.reddit.com/r/REGames/comments/5kp62x/help_reverse_engineering_silent_hill_3_texture/?st=iyn04asn&sh=92f998d9>here</a>
+ *  @note It would seem the 8-bit texture palette comes at the END of the texture, not at beginning like one would expect.
+ *  @note bpp == 32, RGBA; bpp == 24, BGR; bpp == 16, RGBA16; bpp=8, Paletted.
  *
  *  @date 2-1-2017
  *
@@ -29,6 +30,28 @@ namespace sh3_graphics
     /** @defgroup graphics-headers Graphics headers
      *  @{
      */
+
+    #pragma pack(push, 1)
+
+    /**
+     *  Header that comes after the batch header. Contains information about the texture.
+     */
+    struct sh3_texture_info_header
+    {
+        std::uint32_t texHeaderMarker;      /**< Magic Number. This is ALWAYS 0xFFFFFFFF */
+        std::uint32_t unused1;               /**< Unused 32-bit value. Apparently for format identification. */
+        std::uint16_t width;                /**< The hidth of this texture */
+        std::uint16_t height;               /**< The height of this texture */
+        std::uint8_t  bpp;                  /**< Number of bits per pixel of this texture. NOTE: 8bpp is paletted! */
+        std::uint8_t  dataOffset;           /**< #bytes from texHeaderSize+16 to 128 (0 filled) */
+        std::uint16_t padding;              /**< Possibly padding, as it's usually 0 */
+        std::uint32_t texSize;              /**< The size of this texture in bytes (w * h * [bpp/8]) */
+        std::uint32_t texFileSize;          /**< = texSize + texHeaderSize + 16 + endFilleSize */
+        std::uint32_t unknown;              /**< This is unknown/unused */
+        std::uint8_t  widthattrib;          /**< 256 - 8; 32 - 5; 1024 - A; 512 - 9 (I have no clue what this is...) */
+        std::uint8_t  heightattrib;         /**< Same deal */
+        std::uint16_t magic;                /**< Always 0x9999 */
+    };
 
     /**
      * Full texture header.
@@ -59,6 +82,23 @@ namespace sh3_graphics
         std::uint32_t unused7[15];          /**< Unused **/
     };
 
+    /**
+     *  Color palette structure. Contains information about the colour palette.
+     */
+    struct palette_info
+    {
+        std::uint32_t paletteSize;      /**< Number of colors (??) in our color palette */
+        std::uint32_t unused1[2];       /**< Unused as far as I can tell */
+        std::uint8_t  bytes_per_pixel;  /**< Number of bytes per pixel */
+        std::uint8_t  unused2;          /**< Blank Byte */
+        std::uint8_t  entrySize;        /**< Size of one color block in this palette */
+        std::uint8_t  unknown[17];      /**< Unknown or unused bytes */
+        std::uint8_t  distortion;       /**< I have no clue what this is, but it could be important */
+        std::uint8_t  pad[15];          /**< These are all zero. Here so we can align to the palette after a read */
+    };
+
+    #pragma pack(pop)
+
     /**@}*/
 
     /**
@@ -72,6 +112,15 @@ namespace sh3_graphics
     struct sh3_texture final
     {
     public:
+
+        enum PixelFormat
+        {
+            RGBA    = 32,
+            BGR     = 24,
+            RGBA16  = 16,
+            PALETTE = 8,
+        };
+
         sh3_texture(sh3_arc& mft, const std::string& filename){Load(mft, filename);}
         ~sh3_texture(){}
 
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -13,8 +13,7 @@
 #include "SH3/system/window.hpp"
 #include "SH3/system/glprogram.hpp"
 #include "SH3/graphics/msbmp.hpp"
-#include "SH3/graphics/quad.hpp"
-#include "SDL2/SDL.h"
+#include "SH3/graphics/texture.hpp"
 #include <cstdio>
 
 
@@ -33,34 +32,17 @@ int main(int argc, char** argv)
     static_cast<void>(argc);
     static_cast<void>(argv);
 
-    bool quit = false;
+    sh3_arc mft;
+    mft.Load();
 
     Log(LogLevel::INFO, "===SILENT HILL 3 REDUX===");
     Log(LogLevel::INFO, "Copyright 2016-2017 Palm Studios\n");
 
     sh3_window window(640, 480, "sh3redux");
-    sh3_graphics::sh3_glprogram program("test");
 
     sh3_graphics::msbmp("data/pic/error.bmp");
 
-    sh3_graphics::quad rect({vertex3f{-0.5f, 0.5f, 0}, vertex3f{-0.5f, -0.5f, 0}, vertex3f{0.5f, -0.5f, 0}, vertex3f{0.5f, -0.5f, 0}, vertex3f{0.5f, 0.5f, 0}, vertex3f{-0.5f, 0.5f, 0}});
-
-    SDL_Event e;
-
-    glClearColor(1, 0, 0, 1);
-
-    while(!quit)
-    {
-        while(SDL_PollEvent(&e) != 0)
-        {
-            if(e.type == SDL_QUIT)
-                quit = true;
-        }
-
-        glClear(GL_COLOR_BUFFER_BIT);
-        rect.Draw();
-        SDL_GL_SwapWindow(window.hwnd.get());
-    }
+    sh3_graphics::sh3_texture tex(mft, "data/pic/sy/sys_warning.tex");
 
     return static_cast<int>(exit_code::SUCCESS);
 }


source/SH3/arc/vfile.cpp
```diff
@@ -113,7 +113,7 @@ void sh3_arc_vfile::Dump2Disk()
     // Firstly, we need to convert the file path to just the filename.
     std::size_t pos = fname.rfind('/');
 
-    std::ofstream out_file(fname.substr(pos + 1, fname.size()));
+    std::ofstream out_file(fname.substr(pos + 1, fname.size()), std::ios::binary);
```
z33ky: This should be part of another commit.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
 #include <SH3/arc/vfile.hpp>
 
 #include <cassert>
+#include <fstream>
 
+#define HEADER_START 64
```
z33ky: `static const size_t`

This should be moved to the `sh3_texture::Load()` method to give it context.
A better name perhaps, like `BG_HEADER_OFFSET` (as per the current assumption that this is for background images).

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+    std::uint16_t width;
+    std::uint16_t height;
+    std::uint8_t bpp;       // Bits per pixel
+    std::uint8_t flags;
+};
+
+#pragma pack(pop)
+
+/**
+ *  Dump a texture to a Microsoft BMP/DIB
```
z33ky: Isn't it a TARGA (just according to the `tga_header` name)?

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+void Dump2Bitmap(std::uint32_t width, std::uint32_t height, std::vector<std::uint8_t>& data)
+{
+    tga_header header;
+    std::ofstream file("output.tga", std::ios::binary); // Open the stream for binary output
+
+    memset(&header, 0, sizeof(header));
+
+    header.id_size      = 0x00;
+    header.colormap     = 0x00;
+    header.type         = 2;
```
z33ky: Create a named constant for this.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+
+    header.x_origin     = 0x0000;
+    header.y_origin     = 0x0000;
+    header.width        = (uint16_t)width;
+    header.height       = (uint16_t)height;
+    header.bpp          = 24;
+    header.flags        = 0x20; // Origin is upper left hand corner
```
z33ky: Create a named constant for this. An `enum` that is `|`ed together would be even prettier.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+    memset(&header, 0, sizeof(header));
+
+    header.id_size      = 0x00;
+    header.colormap     = 0x00;
+    header.type         = 2;
+
+    header.x_origin     = 0x0000;
+    header.y_origin     = 0x0000;
+    header.width        = (uint16_t)width;
+    header.height       = (uint16_t)height;
```
z33ky: We should `ASSERT` that this cast is safe.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+#pragma pack(pop)
+
+/**
+ *  Dump a texture to a Microsoft BMP/DIB
+ *
+ *  @param width - The Width of this texture.
+ *  @param height - The Height of this texture.
+ *  @param data - Pixel data
+ */
+void Dump2Bitmap(std::uint32_t width, std::uint32_t height, std::vector<std::uint8_t>& data)
```
z33ky: `data` should be a `std::vector<rgb>`.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
 {
     sh3_texture_header          header;
     sh3_arc_vfile               file(mft, filename);
     sh3_arc_vfile::read_error   e;
-    std::vector<std::uint8_t>   data;     // Pixel data of this texture (with the header stripped)
+    std::vector<uint8_t>        data;       // Pixel data of this texture (with the header stripped)
+    std::vector<rgba>           palette;    // Palette Data (I think this is BGRA)
+    std::size_t                 ret;        // How many bytes we get in from file.ReadData( )
 
-    std::size_t ret = file.ReadData(&header, sizeof(header), e); // Read in the header
```
z33ky: I think it made sense to declare `ret` here. This removes the need do document what we use `ret` for above.

In general I'm in favor of the "declare where used paradigm".

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
 #include <SH3/arc/vfile.hpp>
 
 #include <cassert>
+#include <fstream>
 
+#define HEADER_START 64
+
+static std::size_t offset = 0;
```
z33ky: This does not belong in global scope I believe.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
     if(ret != sizeof(header))
-        die("sh3_texture::Load( ): Header only read %d bytes of data! It should have been %d", ret, sizeof(header)); // TODO: Fallback on dummy texture (like Source's purple and black checkboard)
+        die("sh3_texture::Load( ): ret != sizeof(header)!");
 
-    if(header.bpp == 8)
-        Log(LogLevel::WARN, "WARNING: 8-bit textures are currently unsupported! They WILL look like nothing but pixel garbage.");
+    // Check for the pesky 64-byte A7A7A7A7 header that sometimes precedes our texture header
+    if(header.batchHeaderMarker == 0x00000000 && header.batchSize == 0xA7A7A7A7) // AHA!
+    {
+        offset = HEADER_START;
```
z33ky: We could just use a magic number here and justify it's usage.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+    std::vector<uint8_t>        data;       // Pixel data of this texture (with the header stripped)
+    std::vector<rgba>           palette;    // Palette Data (I think this is BGRA)
+    std::size_t                 ret;        // How many bytes we get in from file.ReadData( )
 
-    std::size_t ret = file.ReadData(&header, sizeof(header), e); // Read in the header
+    ret = file.ReadData(&header, sizeof(header), e);
 
     if(ret != sizeof(header))
-        die("sh3_texture::Load( ): Header only read %d bytes of data! It should have been %d", ret, sizeof(header)); // TODO: Fallback on dummy texture (like Source's purple and black checkboard)
+        die("sh3_texture::Load( ): ret != sizeof(header)!");
```
z33ky: I'd find `ReadData() != sizeof(header)` more descriptive.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+    {
+        offset = HEADER_START;
 
-    Log(LogLevel::INFO, "Texture info:\n\ttexWidth:%d\n\ttexHeight:%d\n\ttexSize:%d\n\tbpp:%d", header.texWidth, header.texHeight, header.texSize, header.bpp);
+        file.Seek(offset, std::ios_base::beg);
+        file.ReadData(&header, sizeof(header), e);
 
-    file.Seek(sizeof(header) + header.dataOffset, std::ios_base::beg);
+        if(ret != sizeof(header))
+            die("sh3_texture::Load( ): ret != sizeof(header)!");
```
z33ky: See comment about `die()` above.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+    // Now that we're done that, we can check perform some sanity checks on our texture!
+    if(header.texSize != header.texWidth * header.texHeight * (header.bpp / 8))
+    {
+        Log(LogLevel::WARN, "sh3_texture::Load( ): Warning, texSize != width * height * (bpp / 8)!");
+        return; // TODO: Bind a color shader here
+    }
 
-    if(header.texSize != header.texWidth * header.texHeight * header.bpp)
+    if(header.bpp == PixelFormat::PALETTE)
     {
```
z33ky: This is big. Maybe it deserves its own function.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+
+        // Palette information is stored in blocks (usually of size 64-bytes). We also know how large the
+        // palette is (in bytes, including padding between blocks). From this, we can deduce (with a bit of math)
+        // that the whole palette occupies ~ paletteSize/entrySize bytes of space, contains a total of
+        // entrySize/bypp colors per block, which therefore means we have a total of nBlocks * col_per_block colors,
+        // which equates to about 256-colors in total (which seems accurate for an 8-bit texture).
+
+        std::size_t nBlocks = (pal_header.paletteSize / pal_header.entrySize) / pal_header.bytes_per_pixel; // Number of color blocks
+        std::size_t colors_per_block = pal_header.entrySize / pal_header.bytes_per_pixel;
+        std::size_t count = 0; // Counter
```
z33ky: Unhelpful comment.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+        file.Seek(offset + header.batchHeaderSize + header.texFileSize, std::ios_base::beg);
+        file.ReadData(&pal_header, sizeof(pal_header), e);
+
+        // Palette information is stored in blocks (usually of size 64-bytes). We also know how large the
+        // palette is (in bytes, including padding between blocks). From this, we can deduce (with a bit of math)
+        // that the whole palette occupies ~ paletteSize/entrySize bytes of space, contains a total of
+        // entrySize/bypp colors per block, which therefore means we have a total of nBlocks * col_per_block colors,
+        // which equates to about 256-colors in total (which seems accurate for an 8-bit texture).
+
+        std::size_t nBlocks = (pal_header.paletteSize / pal_header.entrySize) / pal_header.bytes_per_pixel; // Number of color blocks
```
z33ky: Unhelpful comment.

z33ky (commented):
Where I wrote "unhelpful comment", this does not necessarily mean "write a better comment", it could also just be "remove useless comment".

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+    {
+        file.ReadData(&data[0], header.texSize / sizeof(rgba), e);
+    }
+    else if(header.bpp == PixelFormat::BGR)
+    {
+        file.ReadData(&data[0], header.texSize / sizeof(bgr), e);
+    }
+    else if(header.bpp == PixelFormat::RGBA16)
+    {
+        //TODO: Some kind of fucked up shit here. I think this is R5G5B5A1 or something like that..
```
z33ky: `Log(LogLevel::WARN, ...);`

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+            break;
+        case PixelFormat::PALETTE:
+            glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, header.texWidth, header.texHeight, 0, GL_RGB, GL_UNSIGNED_BYTE, &data[0]);
+            break;
+        default:
+            die("sh3_texture::Load( ): Invalid pixel format: %d", header.bpp);
+            break;
+    }
+
+//    glBindTexture(GL_TEXTURE_2D, tex); // Un-bind this texture.
```
z33ky: Why is this commented out?

z33ky (commented):
Clicked on "submit" too soon.

Anyways, I still need to review the palette reading part.

And `main()` just contains whatever we just added now. We could either omit it from the PRs, or (preferably) add a new program (i.e. file containing `main()`) for the thing to test.
This would be pretty easy to do for CMake and a Makefile- or Ninja-based build system, though I imagine this might be cumbersome with IDEs using project files?

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+                            shift++;
+                        }
+
+                        if(shift % 8 == 0 || shift % 8 == 2 || shift % 8 == 4 || shift % 8 == 6)
+                        {
+                            mode |= 2;
+                        }
+                        else
+                        {
+                            mode &= 0xFD;
```
z33ky: Since `mode` is always 0, 1, 2 or 3, this is equivalent to `mode &= 1`.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+        std::uint8_t index = 0;
+        rgba pixel;
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            std::uint32_t tempx = 0;
+            std::uint32_t tempy = 0;
+            std::uint32_t x = 0;
+            std::uint32_t y = 0;
+            std::uint32_t shift = 0;
+            std::uint8_t mode = 2;
```
z33ky: It seems the two mode bits are used separately. This should be representable using two `bool`s.
Until we find a proper name, just call them `mode0` and `mode1`.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+                {
+                    x += 16;
+                }
+                else
+                {
+                    x = 0;
+
+                    if(y + 2 < header.texHeight)
+                    {
+                        if(mode & 1)
```
z33ky: This can be refactored as follows:
```
++y;
if(mode & 1)
{
    y += 2;
}
mode ^= 1;
```

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+                            y++;
+                            mode |= 1;
+                        }
+
+                        if((y % 4) == 0)
+                        {
+                            shift++;
+                        }
+
+                        if(shift % 8 == 0 || shift % 8 == 2 || shift % 8 == 4 || shift % 8 == 6)
```
z33ky: Equivalent to `shift % 2 == 0`, no?

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+                            mode |= 2;
+                        }
+                        else
+                        {
+                            mode &= 0xFD;
+                        }
+                    }
+                    else
+                    {
+                        break;
```
z33ky: Prefer early break, i.e.
```
if(y + 2 >= header.texHeight)
{
    break;
}

if(mode & 1)
...
```

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+                        tempx = x + transformation[i % 32][0];
+                        tempy = y + transformation[i % 32][1];
+                    }
+
+                    iBuffer[(header.texWidth * tempy) + tempx] = index;
+                }
+
+                if(x + 16 < header.texWidth)
+                {
+                    x += 16;
```
z33ky: We could `continue;` here and save one indentation level below.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,330 @@
...
+            }
+        }
+        else // If the distortion flag isn't set, just read the pixel data in from the palette.
+        {
+            for(int i = 0; i < header.texWidth * header.texHeight * 3; i += 3)
+            {
+                file.ReadData(&index, sizeof(index), e);
+                pixel = palette[index];
+
+                data[i]       = pixel.r;
```
z33ky: `i + 0` would create prettier alignment.

source/SH3/graphics/texture.cpp
Quaker762: Forgot to update that sorry.

source/SH3/graphics/texture.cpp
Quaker762: Would you prefer a locally scoped variable or a #define'd value?

source/SH3/graphics/texture.cpp
Quaker762: How would I go about doing that (!= 0 check?)

source/SH3/graphics/texture.cpp
z33ky: I think a class-scoped static constant is good.

Since many of these fields have sane default values, you could also use member initialization in the class definition like so:
```
struct tga_header
{
    static constexpr std::uint8_t TYPE_RGB24 = 2;

    std::uint8_t id_size = 0;
    std::uint8_t colormap = 0;
    std::uint8_t type = TYPE_RGB24;
    ...
};
```
So we don't need the `memset` or many of the assignments here.

We could also list all possible types in an enum, but if we're only going to use RGB24 anyway it's not too useful. I wouldn't necessarily be against it though.

source/SH3/graphics/texture.cpp
z33ky: `ASSERT(width <= std::numeric_limits<decltype(header.width)>::max());`

I might be overly pedantic about integer type casts...

source/SH3/graphics/texture.cpp
Quaker762: The only issue with this is the mismatch between the palette and data. We want an `rgb24` image, but the palette is a 32-bit pixel (the alpha channel is always `0x80` for some weird reason..)

source/SH3/graphics/texture.cpp
Quaker762: How do you mean? Just set offset to 0x40 without worrying about a constant and just comment it?

source/SH3/graphics/texture.cpp
Quaker762: Probably. I just thought it was easier as we already had all the data we needed without having to call anything else.

source/SH3/graphics/texture.cpp
Quaker762: This was sort of just for my own sanity haha.

source/SH3/graphics/texture.cpp
Quaker762: I commented the whole function before I rewrote it. This missed my selection before uncomment haha.

source/SH3/graphics/texture.cpp
Quaker762: I still need to `memset` the unused bytes otherwise they get filled with crap and paint.net won't open it anymore :\

z33ky: You won't if you use member initialization for *every* member in the struct.

source/SH3/graphics/texture.cpp
z33ky: Yeah.

Quaker762:
I seem to have fixed the slight offset in the x-axis in the texture (tempx - 16), however there is still a fair amount of pixel noise, though I'm not sure where it's coming from. Something must be a bit screwy with the matricies, or the palette is also encrypted (which I seriously doubt)

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+
+struct tga_header
+{
+    static constexpr std::uint8_t TYPE_RGB24 = 2;
+    static constexpr std::uint8_t FLAGS_FLIP = 0x20;
+
+    std::uint8_t id_size = 0;               /**< Size of the ID field that follows this header (usually 0) */
+    std::uint8_t colormap = 0;              /**< Is this image paletted (THIS MUST ALWAYS BE 0 FOR US!) */
+    std::uint8_t type = TYPE_RGB24;         /**< Type of image (ALWAYS 2 FOR RGB24!) */
+    std::uint8_t unused[5] = {0, 0, 0, 0, 0};
```
z33ky: You can just `= {};`.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            std::uint32_t tempx = 0;
+            std::uint32_t tempy = 0;
+            std::uint32_t x = 0;
+            std::uint32_t y = 0;
+            std::uint32_t shift = 0;
+            //std::uint8_t mode = 2;
+
+            bool mode0 = false;;
```
z33ky: `;;` -> `;`

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+            bool mode1 = true;
+
+            // I don't really know what these are used for, the original code has no explanation.
+            // I think it has something to do with re-ordering the indicies.
+            unsigned char normal[][2] =
```
z33ky: These can be `static constexpr`.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+            };
+
+            // We need to step through each index and unencrypt it
+            for(int i = 0; i < header.texSize;)
+            {
+                for(int j = 0; j < 32; j++, i++)
+                {
+                    file.ReadData(&index, sizeof(index), e);
+
+                    if(!(mode1))
```
z33ky: `(mode1)` -> `mode1`

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+        data.resize(header.texWidth * header.texHeight * 3);
+        iBuffer.resize(header.texSize);
+
+        file.Seek(offset + (header.texFileSize - header.texSize), std::ios_base::beg); // Seek to the beginning of data
+
+        std::uint8_t index = 0;
+        rgba pixel;
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            std::uint32_t tempx = 0;
```
z33ky: The `temp{x,y}` variables can be declared in the loop below, reducing their scope.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
 
-    file.Seek(sizeof(header) + header.dataOffset, std::ios_base::beg);
+                x = 0;
+                if(y + 2 >= header.texHeight)
+                {
+                    break;
+                }
 
-    data.resize(header.texSize);
+                if(mode0)
```
z33ky: If you don't want the structure I was proposing last iteration (totally fine, since we don't have an intuitive understanding of the algorithm yet) I would still pull the `mode0` toggle outside the branches and just do `mode0 = !mode0`.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+                {
+                    y++;
+                    //mode |= 1;
+                    mode0 = true;
+                }
 
-    if(header.texSize != header.texWidth * header.texHeight * header.bpp)
+                if((y % 4) == 0)
+                {
+                    shift++;
```
z33ky: Every time `shift` is even, `mode1` is set to `true` below, if it is uneven it is set to `false`.
It looks to me like we can remove `shift` entirely and just do `mode1 = !mode1` in here.

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+}
+
 //TODO: Scale the texture and then
 void sh3_texture::Load(sh3_arc& mft, const std::string& filename)
 {
     sh3_texture_header          header;
     sh3_arc_vfile               file(mft, filename);
     sh3_arc_vfile::read_error   e;
-    std::vector<std::uint8_t>   data;     // Pixel data of this texture (with the header stripped)
+    std::vector<uint8_t>        data;       // Pixel data of this texture (with the header stripped)
```
z33ky: `std::uint8_t`

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
+ *  @param width - The Width of this texture.
+ *  @param height - The Height of this texture.
+ *  @param data - Pixel data
+ */
+void Dump2Bitmap(std::uint32_t width, std::uint32_t height, std::vector<std::uint8_t>& data)
+{
+    tga_header header;
+    std::ofstream file("output.tga", std::ios::binary); // Open the stream for binary output
+
+    header.width        = (uint16_t)width;
```
z33ky: `std::uint16_t`

source/SH3/graphics/texture.cpp
```diff
@@ -13,48 +13,329 @@
...
 
+#pragma pack(push, 1)
+struct rgba
+{
+    std::uint8_t r, g, b, a;
+};
+
+struct rgba16
+{
+    uint8_t r : 5;
```
z33ky: `std::uint8_t`

z33ky (changes requested):
We could think about making `data` a `union{ std::vector<rgb> rgb; std::vector<rgba> rgba; ...}`.
With this we can remove the loops reading the pixels for `PixelFormat::PALETTE` with `std::transform`.

z33ky:
> I seem to have fixed the slight offset in the x-axis in the texture (tempx - 16)

This can't be right. Starting with `x = 0, y = 0, i = 0`, `tempx` will be `-12` and `tempy` will be `0`. Accessing `iBuffer` with the index `(header.texWidth * tempy) + tempx = -12` is out of range.

Perhaps it's got something to do with the check in line 228, or the addition in 230?

source/SH3/graphics/texture.cpp
Quaker762: Isn't is bad to declare variables in a loop, or is the compiler smart enough to not create a massive stack.

z33ky: The compiler will definitively not increase the size of the stack on each iteration for this.
If we were unnecessarily invoking a constructor, or computing some value to initialize the variable on each iteration, then performance would suffer.

Quaker762:
> This can't be right. Starting with x = 0, y = 0, i = 0, tempx will be -12 and tempy will be 0. Accessing iBuffer with the index (header.texWidth * tempy) + tempx = -12 is out of range.

I thought this as well, but see for yourself (bottom left corner of both images)

img1

img2

> Perhaps it's got something to do with the check in line 228, or the addition in 230?

I'm not sure, but it works. I have no clue how the actual encryption works, but I suspect it does some kind of block translation. I also think the matrices are slightly off, which is causing the shift/noisy output on some colours.

source/SH3/graphics/texture.cpp
Quaker762: Sorry I'm a bit confused, what do you want me to do here? Some of this stuff is REALLY fragile and it crashes if I make really large changes.

source/SH3/graphics/texture.cpp
Quaker762: If I replace `shift++` with `mode1 = !mode1` and remove the `if/else` branch, the image comes out distorted.

source/SH3/graphics/texture.cpp
z33ky: See [this](https://github.com/Palm-Studios/sh3redux/pull/87#discussion_r102255094).
Alternatively,
```
if(mode0)
{
    y += 3;
    mode1 = true;
}
else
{
    y++;
}
mode0 = !mode0;
```

source/SH3/graphics/texture.cpp
z33ky: Was the `mode1 = true` in line 224 always there? If not, this throws things off.

z33ky:
> I thought this as well, but see for yourself

That doesn't stop it from being an out-of-bounds access. It works that way *now*, but relying on undefined behavior has seldom treated projects well.
To me, the `- 16` crutch rather indicates that some code causes `x` to be greater than it ought to be, perhaps even only in some cases, which might account for some of the noise.

I'll take a stab at it tomorrow.

Quaker762:
> To me, the - 16 crutch rather indicates that some code causes x to be greater than it ought to be, perhaps even only in some cases, which might account for some of the noise

I thought this may have been the case as well, but I got some rather strange output. I'll email the guy and see if he can shed some light, it's a shame he didnt comment his code :/

source/SH3/graphics/texture.cpp
Quaker762: I'm not sure. I'm a bit tured so I'll go over this tomorrow when I'm thinking right.

Quaker762:
Also wouldn't an out of bounds read cause a segfault? Or does the STL have measures to stop that.

z33ky:
> Also wouldn't an out of bounds read cause a segfault?

Assuming nothing funky happens due to the undefinedness of behavior, only when the access happens on an address that belongs to a page (chunk of memory, usually 4KiB on x86) not mapped to the process.

> Or does the STL have measures to stop that.

`operator[]` generally does not do bounds-checking. The `at()` methods does however, which can thus prevent a segfault - assuming the STL container is in a valid state.
You can also try the address sanitizer.

Quaker762:
Don't merge yet I'm going to do some stuff for 16 bit textures.

Quaker762:
I've just had to add a slight hack, it looks like Konami is at it again... There is a mismatch between 24 and 32-bit, it would seem all the 32-bit textures are labelled as being 24-bit....

wot

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,326 @@
...
+
+
+    glGenTextures(1, &tex);             // Create a texture
+    glBindTexture(GL_TEXTURE_2D, tex);  // Bind it for use
+
+    // Create the texture according to its pixel format!
+    switch(header.bpp)
+    {
+        case PixelFormat::RGBA:     // Regular 32-bit RGBA
+            glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, header.texWidth, header.texHeight, 0, GL_RGBA, GL_UNSIGNED_BYTE, &data[0]);
```
z33ky: We can replace these with
```
GLenum srcFormat;
GLint dstFormat;
switch(header.bpp)
{
    case PixelFormat::RGBA:
        srcFormat = GL_RGBA;
        dstFormat = GL_RGBA;
        break;
    ...
}
glTexImage2D(GL_TEXTURE_2D, 0, dstFromat, ..., srcFormat, ...);
```

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,326 @@
...
+            break;
+        case PixelFormat::BGR:      // 24-bit BGR
+            glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, header.texWidth, header.texHeight, 0, GL_BGR, GL_UNSIGNED_BYTE, &data[0]);
```
z33ky: I think you sapped `internalFormat` and `format` here and in the following calls.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,326 @@
...
+            break;
+        case PixelFormat::RGBA16:   // 16-bit RGBA. OpenGL supports this (I think)
```
z33ky: It does, but differently:
`glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, ..., GL_RGBA, GL_UNSIGNED_SHORT_5_5_5_1, ...);`

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,326 @@
...
+
 //TODO: Scale the texture and then
 void sh3_texture::Load(sh3_arc& mft, const std::string& filename)
 {
     sh3_texture_header          header;
     sh3_arc_vfile               file(mft, filename);
     sh3_arc_vfile::read_error   e;
-    std::vector<std::uint8_t>   data;     // Pixel data of this texture (with the header stripped)
+    std::vector<uint8_t>        data;       // Pixel data of this texture (with the header stripped)
+    std::vector<bgra>           palette;    // Palette Data (I think this is BGRA)
```
z33ky: This should be moved down to the palette branch.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,326 @@
...
+        }
+
+        // Now that we've completely loaded the palette in its entirety, we can get the 8-bit index value
+        // from the data section of the file and get it's color in the palette!
+
+        //===---THIS IS A CLUSTER FUCK FOR NOW UNTIL WE UNDERSTAND HOW IN THE NAME OF CHRIST THIS WORKS---===//
+        //std::vector<std::uint8_t> iBuffer;  // Our index buffer that we put transformed indecies into
+        std::vector<std::int16_t> iBuffer;  // Our index buffer that we put transformed indecies into
+
+        data.resize(header.texWidth * header.texHeight * 3);
```
z33ky: We should put a comment that we expect the palette to not contain transparent colors, even though it's BGRA.

z33ky (changes requested):

Quaker762:
@z33ky  Whoa, for some reason after my rebase it's pushed everything... I haven't done anything differently than usual....?? Should I do a rebase/reset??

z33ky:
Seems the rebase didn't do right. Make sure your remotes are up-to-date (`git fetch --all`) and try again.
If it still doesn't work, try cherry-picking this commit onto master.

Quaker762:
What do I run after `git fetch --all`? It's probably going to be easier to cherry pick if you're happy with everything.

z33ky:
Try the rebase again (`git rebase upstream/master`).
For cherry-picking, do `git branch --move texture_new texture_new_old && git checkout -b texture_new upstream/master && git cherry-pick texture_new_old`.

Quaker762:
All fixed

source/SH3/graphics/texture.cpp
z33ky: Still unchanged.

source/SH3/graphics/texture.cpp
z33ky: Still unchanged.

source/SH3/graphics/texture.cpp
z33ky: We'll want to rename it to `DumpRGB2Bitmap` at least, to make sure it's not used for RGBA.
Also this function can be an an anonymous namespace.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+};
+
+#pragma pack(pop)
+
+/**
+ *  Dump a texture to a TARGA/TGA file
+ *
+ *  @param width - The Width of this texture.
+ *  @param height - The Height of this texture.
+ *  @param data - Pixel data
```
z33ky, line 80: This should specify the format.

source/SH3/graphics/texture.cpp
z33ky: Still unchanged.
Also, old-style-cast (should use `staitc_cast` instead).

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+ */
+void Dump2Bitmap(std::uint32_t width, std::uint32_t height, std::vector<std::uint8_t>& data)
+{
+    tga_header header;
+    std::ofstream file("output.tga", std::ios::binary); // Open the stream for binary output
+
+    header.width        = (uint16_t)width;
+    header.height       = (uint16_t)height;
+
+    assert(width <= std::numeric_limits<decltype(header.width)>::max());
```
z33ky: This check should happen before casting.
Also, use `ASSERT` :)

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+    assert(height <= std::numeric_limits<decltype(header.height)>::max());
+
+    file.write((char*)&header, sizeof(header));
```
z33ky: old-style-cast

source/SH3/graphics/texture.cpp
z33ky: Still unchanged.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
 void sh3_texture::Load(sh3_arc& mft, const std::string& filename)
 {
     sh3_texture_header          header;
     sh3_arc_vfile               file(mft, filename);
     sh3_arc_vfile::read_error   e;
-    std::vector<std::uint8_t>   data;     // Pixel data of this texture (with the header stripped)
+    std::vector<uint8_t>        data;       // Pixel data of this texture (with the header stripped)
 
-    std::size_t ret = file.ReadData(&header, sizeof(header), e); // Read in the header
+    static std::size_t          offset = 0;
```
z33ky: Doesn't need to be `static`.
Should be of type `std::streamsize`.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+        file.Seek(offset, std::ios_base::beg);
+        file.ReadData(&header, sizeof(header), e);
 
-    file.Seek(sizeof(header) + header.dataOffset, std::ios_base::beg);
+        if(ret != sizeof(header))
+            die("sh3_texture::Load( ): ReadData( ) != sizeof(header)!");
+    }
 
-    data.resize(header.texSize);
+    if(header.texSize == (header.texWidth * header.texHeight* 4))
```
z33ky: Unnecessary parenthesis.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+    {
+        header.bpp = 32; // Thanks KONAMI!
+    }
 
-    if(header.texSize != header.texWidth * header.texHeight * header.bpp)
+    // Now that we're done that, we can check perform some sanity checks on our texture!
+    if(header.texSize != header.texWidth * header.texHeight * (header.bpp / 8))
```
z33ky: Unnecessary parenthesis.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+        std::vector<bgra>    palette;    // Palette Data (I think this is BGRA)
 
-    glBindTexture(GL_TEXTURE_2D, tex);
-    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, header.texWidth, header.texHeight, 0, GL_BGRA, GL_UNSIGNED_BYTE, &data[0]);
-    glBindTexture(GL_TEXTURE_2D, 0); // Unbind this texture
+        // First, we need to seek to the palette and read it in.
+        file.Seek(offset + header.batchHeaderSize + header.texFileSize, std::ios_base::beg);
+        file.ReadData(&pal_header, sizeof(pal_header), e);
+
+        ASSERT(pal_header.bytes_per_pixel == sizeof(palette.front()));
```
z33ky: Should always check (not just in debug builds) and return if check fails.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+
+        const std::size_t nBlocks = (pal_header.paletteSize / pal_header.entrySize) / pal_header.bytes_per_pixel;
+        const std::size_t colorsPerBlock = pal_header.entrySize / pal_header.bytes_per_pixel;
+
+        palette.resize(colorsPerBlock * nBlocks);
+
+        for(std::size_t block = 0; block < nBlocks; ++block)
+        {
+            std::size_t read = file.ReadData(static_cast<void*>(&palette[block * colorsPerBlock]), pal_header.entrySize, e);
+            ASSERT(read == pal_header.entrySize);
```
z33ky: Should always check and warn if check fails.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+            // Each block is 256 bytes.
+            // Seek to next block.
+            file.Seek(256 - pal_header.entrySize, std::ios_base::cur);
+        }
+
+        // we need to swap some colors every 32 pixels, starting from the 8th
+        static constexpr std::size_t swapDistance = 32; // distance between swaps
+        static constexpr std::size_t swapSize = 8; // amount of colors to swap
+        ASSERT(palette.size() > 8);
```
z33ky: Should always check and warn and skip the loop below if check fails.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+            }
+            std::swap_ranges(iter, swapBlock, swapBlock);
+        }
+
+        // Now that we've completely loaded the palette in its entirety, we can get the 8-bit index value
+        // from the data section of the file and get it's color in the palette!
+
+        //===---THIS IS A CLUSTER FUCK FOR NOW UNTIL WE UNDERSTAND HOW IN THE NAME OF CHRIST THIS WORKS---===//
+        //std::vector<std::uint8_t> iBuffer;  // Our index buffer that we put transformed indecies into
+        std::vector<std::int16_t> iBuffer;  // Our index buffer that we put transformed indecies into
```
z33ky: Change back to `std::uint8_t`.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+
+        data.resize(header.texWidth * header.texHeight * 3); // We strip the Alpha channel from the BGRA pixel beacuse it is hard locked to 0x80 (not 0xFF!!)
+        iBuffer.resize(header.texSize, -1); // FIXME: -1 filling just for debugging
```
z33ky: Remove `-1` filling.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+
+        file.Seek(offset + (header.texFileSize - header.texSize), std::ios_base::beg); // Seek to the beginning of data
+
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            ASSERT(header.texWidth % 16 == 0);
```
z33ky: Should always check (?). Do something smart if it fails.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+                    if(offsetFlipper)
+                    {
+                        xoffset ^= 4;
+                    }
+
+                    const auto tempx = x + xoffset; // - 16;
+                    // every other pixel is for (y + 2)
+                    const auto tempy = y + ((i % 2) ? 2 : 0);
+
+                    ASSERT(iBuffer[(header.texWidth * tempy) + tempx % header.texWidth] == -1);
```
z33ky: Remove.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            ASSERT(header.texWidth % 16 == 0);
+
+            std::uint32_t x = 0;
+            std::uint32_t y = 0;
+            bool offsetFlipper = false;
+
+            // We need to step through each index and unencrypt it
+            while(true)
```
z33ky: `//FIXME: distortion on 16-pixel wide block on the left`

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+                x = 0;
+
+                ++y;
+                if(y % 2 == 0)
+                {
+                    // each iteration we read two lines: offset 0 and 2
+                    // so after two iterations, we have read the lines with offset 0 and 2, 1 and 3
+                    // so now we need to skip two additional lines, since we read 2 and 3 already
+                    y += 2;
+                    ASSERT(y <= header.texHeight);
```
z33ky: Always check, warn & `break` if it fails.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,340 @@
...
+                if(y == header.texHeight)
+                {
+                    break;
+                }
+            }
+
+            // Christ help me...
+            for(std::size_t pindex = 0, doff = 0; pindex < iBuffer.size() && doff < data.size(); pindex++, doff += 3)
+            {
+                ASSERT(iBuffer[pindex] != -1);
```
z33ky: Remove.

z33ky (changes requested):
There's also some `sign-compare` and `sign-conversion` warnings that should be fixed, in most cases by simply explicitly casting, perhaps with previous checking if the casts are safe.

source/SH3/graphics/texture.cpp
Quaker762: As in pass in the bpp of the texture?

source/SH3/graphics/texture.cpp
Quaker762: Is `ASSERT` for debug builds? I ahven't really checked out the `assert` module we have.

source/SH3/graphics/texture.cpp
Quaker762: Would you rather I use an `if/else` or a label (the C is leaking back in haha :P)

source/SH3/graphics/texture.cpp
z33ky: I actually meant further qualifying "pixel data" in the comment.
Passing the format (RGB8, RGB5A1, ...) would also work very well, though it would require the function to be extended to support other formats - or simply refuse to work with non-RGB formats.

source/SH3/graphics/texture.cpp
z33ky: Both `ASSERT` and `assert` are only enabled in debug mode by default. There's some preprocessor flags to control the behavior, which for `ASSERT` is also directly exposed in the CMake configuration.
Also, didn't you review the  `assert.hpp` commit?

source/SH3/graphics/texture.cpp
z33ky: I'd prefer the if/else. It doesn't add complexity vs `goto` and visually highlights that the loop is conditional with another indentation level.

source/SH3/graphics/texture.cpp
Quaker762: It was a while ago sorry, I'll look over it again.

source/SH3/graphics/texture.cpp
Quaker762: Not really much we can do except fail. The textures are all power of 2 IIRC.

source/SH3/graphics/texture.cpp
z33ky: In most languages, `assert`s are a feature used for debugging, although it can usually be controlled via a separate flag. For the `assert` in the standard library, once has to define the `NDEBUG` macro to get rid of them, which CMake does for the release configuration.

As a side-note, since `assert`s are removed from release-builds, I'm not sure if the optimizer can make the same assumptions. We might want to modify `ASSERT` to do `if(!x) __builtin_unreachable()` when disabled.

source/SH3/graphics/texture.cpp
Quaker762: I could pass in the bitness and then just change the value in the header. I think TGA stores everything as R5G5B5A1 anyway, but yeah, it would be reversed.

z33ky: Then perhaps just rename to `DumpRGB2Bitmap` to indicate the order of colors and include the bpp parameter like you said.

Quaker762:
I just realised I made a typo on `line 103` where I stated the default is 32 not 24

EDIT: I just broke it, it only dumps black images. Looking for the cause now.

Alright it's being caused by `if(y <= header.texHeight)` on `line 278`.. I had it reversed haha.

source/SH3/graphics/texture.cpp
z33ky: We can truncate (`header.texWidth -= header.texWidth % 16`) and warn.
I think `header.texHeight % 4 == 0` also has to hold, so we should to the same for height.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,380 @@
...
+{
+/**
+ *  Dump a texture to a TARGA/TGA file
+ *
+ *  @param width - The Width of this texture.
+ *  @param height - The Height of this texture.
+ *  @param data - Pixel data
+ *  @param bpp - Bitness of the texture to dump.
+ */
+void DumpRGB2Bitmap(std::uint32_t width, std::uint32_t height, std::vector<std::uint8_t>& data, std::uint8_t bpp = 24)
```
z33ky: I'd be careful with selecting a default here. I'd make it a required parameter to ensure correct usage.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,380 @@
...
+    switch(bpp)
+    {
+    case 16:
+        header.bpp = 16;
+        break;
+    case 32:
+        header.bpp = 32;
+        break;
+    default:
+        break; // Do nothing as we have a default value of 32 (RGB).
```
z33ky: I'd include `24` and use the `default` label to check that `bpp` isn't something weird.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,380 @@
...
+                std::swap_ranges(iter, swapBlock, swapBlock);
+            }
+        }
+
+
+        // Now that we've completely loaded the palette in its entirety, we can get the 8-bit index value
+        // from the data section of the file and get it's color in the palette!
+
+        //===---THIS IS A CLUSTER FUCK FOR NOW UNTIL WE UNDERSTAND HOW IN THE NAME OF CHRIST THIS WORKS---===//
+        //std::vector<std::uint8_t> iBuffer;  // Our index buffer that we put transformed indecies into
```
z33ky: remove

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,380 @@
...
+            {
+
+            }
+
+            std::uint32_t x = 0;
+            std::uint32_t y = 0;
+            bool offsetFlipper = false;
+
+            // We need to step through each index and unencrypt it
+            while(true) // FIXME: distortion on 16-pixel wide block on the left
```
z33ky: I think this makes it look like the comment belongs to the looping condition, while (heh) it refers to the whole block.
I'd put it in a separate line above.

source/SH3/graphics/texture.cpp
z33ky: This `ASSERT` can be removed now.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,380 @@
...
+        for(std::size_t block = 0; block < nBlocks; ++block)
+        {
+            std::size_t read = file.ReadData(static_cast<void*>(&palette[block * colorsPerBlock]), pal_header.entrySize, e);
+
+            if(read != pal_header.entrySize)
+            {
+                Log(LogLevel::WARN, "sh3_texture::Load( ): Warning: Number of bytes read in palette block != pal_header.entrySize! (Expected %d, got %d)", read, pal_header.entrySize);
+            }
+
+            ASSERT(read == pal_header.entrySize);
```
z33ky: There's nothing really useful for us to debug if this fails. I don't think we need this anymore.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,380 @@
...
+                    // so after two iterations, we have read the lines with offset 0 and 2, 1 and 3
+                    // so now we need to skip two additional lines, since we read 2 and 3 already
+                    y += 2;
+
+                    if(y <= header.texHeight)
+                    {
+                        Log(LogLevel::WARN, "sh3_texture::Load( ): Warning: y <= header.texHeight!");
+                        break;
+                    }
+                    ASSERT(y <= header.texHeight);
```
z33ky: Same as other `ASSERT`. Remove.

z33ky (commented):
I inserted the `ASSERTS` to make sure I didn't break anything when changing stuff. Most of them were never intended to make it into the "real" commit, at least not as is.

source/SH3/graphics/texture.cpp
z33ky: The header probably shouldn't set a default `bpp` either. This also removed the special-casing for `24`.

Quaker762: Sure thing. It does make more sense to not have a default, I just assumed for the most part that most textures won't have an alpha channel.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,385 @@
...
+
+    ASSERT(width <= std::numeric_limits<decltype(header.width)>::max());
+    ASSERT(height <= std::numeric_limits<decltype(header.height)>::max());
+
+    header.width        = static_cast<std::uint16_t>(width);
+    header.height       = static_cast<std::uint16_t>(height);
+
+    switch(bpp)
+    {
+    case 8:
```
z33ky: `bpp` should specify the bbp for `data`, not the original image.

source/SH3/graphics/texture.cpp
```diff
@@ -9,52 +9,382 @@
...
+            std::size_t read = file.ReadData(static_cast<void*>(&palette[block * colorsPerBlock]), pal_header.entrySize, e);
+
+            if(read != pal_header.entrySize)
+            {
+                Log(LogLevel::WARN, "sh3_texture::Load( ): Warning: Number of bytes read in palette block != pal_header.entrySize! (Expected %d, got %d)", read, pal_header.entrySize);
+            }
+
+            // Each block is 256 bytes.
+            // Seek to next block.
+            file.Seek(256 - pal_header.entrySize, std::ios_base::cur);
```
z33ky, line 190: I'm unsure if we maybe just want to break on the condition above, since I wouldn't expect to read any more bytes if it failed once.
If we continue this should seek to `256 - read`.

z33ky (changes requested):
Are you still working on this? Because there's still some `ASSERT` stuff to be resolved (see yesterdays reviews).

Quaker762:
@z33ky Sorry the semesters just started up and I haven't had a lot of time to sit down and comb through it all. I've only got one class tomorrow so I'll get on it ASAP.

source/SH3/graphics/texture.cpp
Quaker762: So you're suggesting we just leave the loop early then? Should we perhaps let the call to Load () be handled by the calling function instead of the constructor (and have Load () return an error code) so we can just bind a flat colour/checkboard shader?

z33ky:
Hey man, with how much code I've wrote I'm not complaining ;o

source/SH3/graphics/texture.cpp
Quaker762: Or perhaps we should leave that idea until we come to it (on second thought)

Quaker762:
Hahaha no worries, you're making me a much better programmer I can tell you that haha. Hopefully you can make some more PRs when the ball really gets rolling and we're not twiddling with these silly formats. :)

source/SH3/graphics/texture.cpp
```diff
@@ -9,57 +9,383 @@
...
+        data.resize(header.texWidth * header.texHeight * 3); // We strip the Alpha channel from the BGRA pixel beacuse it is hard locked to 0x80 (not 0xFF!!)
+        iBuffer.resize(header.texSize);
+
+        file.Seek(offset + (header.texFileSize - header.texSize), std::ios_base::beg); // Seek to the beginning of data
+
+        if(header.texWidth > 96) // Apparently this is the distortion flag?!?!
+        {
+            if(header.texWidth % 16 != 0)
+            {
+
```
z33ky: Warn, `break`.

source/SH3/graphics/texture.cpp
```diff
@@ -9,57 +9,383 @@
...
+            }
+
```
z33ky: Also check `header.texHeight % 4`.

z33ky (changes requested):
I also just noticed that the commit message is in past tense. It should use present tense.

source/SH3/graphics/texture.cpp
z33ky: Err, by `break` I mean truncate.

Quaker762: Truncate?

source/SH3/graphics/texture.cpp
Quaker762: Sure thing

source/SH3/graphics/texture.cpp
z33ky: `header.texWidth -= header.texWidth % 16` to align it to a 16-pixel boundary.

Edit:
I'm unsure if we maybe want to keep the original texture size (i.e. make a local variable `texWidth = header.texWidth - header.texWidth % 16` that is used in the loop below).
But since we don't expect this to happen anyways I think truncation is fine.

Quaker762: It's probably a good idea to hang onto some of the txture metadata, but I don't think it's necessary at this point. Maybe if we realise we need it later. 

> But since we don't expect this to happen anyways I think truncation is fine.

Yeah, I honestly doubt we'll run into this, otherwise SH3 wouldn't have worked in the first place haha.

z33ky: > Yeah, I honestly doubt we'll run into this, otherwise SH3 wouldn't have worked in the first place haha.

Yah, I'm overly cautious about making assumptions.
If the data is corrupted it's highly probable other errors will make it fail anyways. Failing gracefully is pretty, but not really useful in this case.
And I doubt there's any mods that tried to mess with these paletted textures.

z33ky (approved):

z33ky:
Just `commit --amend` that message into present tense and you can merge it.

z33ky:
Just `commit --amend` the message into present tense and you can merge it.

PR #90, opened 2017-03-31
Make Heather happy

z33ky:
Heather looks kind of distraught in the icon, but I think that she should be happy, because we're revamping the game.
I also changed the background to a soothing sky-blue :blue_heart:

Perhaps we should also replace the textures in-game to make everyone look happy :smiley:

The game is also very dark in many areas, so dark, that Heather needs to use a flashlight to see properly. I think we should light up the place and clean the textures, to make the world look more welcoming :high_brightness:


z33ky:
24h later I suddenly realized this is **fool**ish.

Closing to keep the world dark and scary.

Quaker762:
HAHAHAH absolutely brilliant PR. This should actually be merged, silent hill is a bit of a scary place, which is the one thing Konami screwed up on. 

Approved and merged ;)

PR #93, opened 2017-06-05
Slight tweak to sh3_glcontext error handling

Quaker762:
GLEW init failure now uses die(...) instead of just logging, aborting and returning -1 (which would most certainly confuse a user that isn't a programmer).


Quaker762:
As a side, @z33ky how is your thesis coming along?

z33ky:
I'm starting to prepare for my defense. So yeah, it's going well :)

PR #94, opened 2017-06-14
Abstraction of OpenGL Data types

Quaker762:
Very basic abstraction of more complex OpenGL concepts (Vertex Array Objects and Vertex Buffer Objects). Makes the whole thing _much_ more readable and easier to handle. Should also make everything a little bit
more typesafe and memory safe (no segfaults).

There's also a nice demo in main as to how it works. As is clearly visible, there is a LOT less OpenGL bullshit to type out, as the object does it all for you! :)


Quaker762:
Oh, and I've discovered possibly the best thing ever. It's called CodeXL (it's an OpenGL/OpenCL/CPU Debugger). I'll update the readme, as _everyone_ should be using this to debug 3D data (especially when we get to anything cutscene releated).

debug

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include <cstddef>
+
+namespace sh3_gl
+{
+    /**
+     *
+     * An OpenGL Vertex Buffer Object (VBO). Provides a nice, easy, typesafe way
+     * to create and manage OpenGL buffers.
+     */
+    struct sh3_glbuffer final
```
z33ky: The name seems redundant with the namespace. I think this should just be `buffer`.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+    public:
+
+        enum Target
```
z33ky: Any reason why this is not an `enum class`?

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+            SH3_ELEMENT_BUFFER  = GL_ELEMENT_ARRAY_BUFFER,
+            SH3_TEXTURE_BUFFER  = GL_TEXTURE_BUFFER
+        };
+
+        /**
+         *  Default constructor.
+         */
+        sh3_glbuffer(){Create();buffType = Target::SH3_ARRAY_BUFFER;};
+        sh3_glbuffer(Target type){Create();buffType = type;};
+        sh3_glbuffer(Target type, const std::string& buffname){Create();buffType = type, name = buffname;};
```
z33ky: You could provide default values for those parameters to get rid of the overloads.
Also, is it important that `Create()` is called before initializing the members? Otherwise you can use initializer lists.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+        /**
+         *  Destructor. Ensure we have no leaks when this buffer is to be deleted by the C++ Runtime.
```
z33ky: "by the C++ Runtime"??
Anyway, seems like an redundant statement, since "Destructor" should imply as much.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+         */
+        ~sh3_glbuffer(){/**Release();*/};
```
z33ky: Why is this commented out?

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+         */
+        void Release();
+
+        /**
+         *  Give some data to our buffer
+         *
+         *  @param num Size (in bytes) of the data we are flushing.
+         *  @param data void pointer to our data (usually an std::array or std::vector).
+         */
+        void BufferData(std::size_t num, void* data);
```
z33ky: I didn't actually check if this is indeed common, but I think most APIs pass the data first and then the length. I'd like to stay consistent if there's a rule.
Also, I think it'd make sense to provide an API like `sh3_arc_file::ReadObject` which uses templates to infer the size.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,112 @@
...
+         */
+        Target GetType() const {return buffType;}
+
+        /**
+         *  Get the number of elements contained within this buffer.
+         *
+         *  @return Number of elements in this GL Buffer.
+         */
+        std::uint32_t GetNumberElements() const;
+        std::string     name; /**< The name of this buffer */
```
z33ky: Any reason to have this not be `const`?

include/SH3/system/glprogram.hpp
```diff
@@ -48,9 +48,11 @@ namespace sh3_graphics
         private:
         };
 
+        sh3_glprogram(const std::string& name) : programName(name){Load(name);}
```
z33ky: I'm not sure if I'm a fan of this overload. The version below still allows to ignore errors, but makes it visible to the caller that there could be one that can be checked for.

include/SH3/system/glprogram.hpp
```diff
@@ -48,9 +48,11 @@ namespace sh3_graphics
...
         sh3_glprogram(const std::string& name, load_error& err, const std::vector<std::string>& attribs = {}) : programName(name){Load(name, err, attribs);};
         ~sh3_glprogram(){Unbind(); glDeleteProgram(programID);}
 
+        void Load(const std::string& name);
```
z33ky: See above.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,118 @@
...
+    public:
+
+        sh3_glvertarray(){Create();};
+
+        /**
+         *  VAO Attribute locations. These are used in a @ref sh3_glprogram to access data
+         *  sent to it from a VBO. These values were picked for no specific reason other than
+         *  the usual order of a vertex (vert data, normal data, uv data).
+         */
+        enum Attribute
```
z33ky: Any reason why this is not an `enum class`?

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,118 @@
...
+            SH3_VERTEX_SLOT = 0,
+            SH3_NORMAL_SLOT = 1,
+            SH3_UV_SLOT = 2,
+            SH3_INDEX_SLOT = 3
+        };
+
+        /**
+         *  Specifies data type of data passed to paramater 3 of glVertexAttribPointer.
+         */
+        enum DataType
```
z33ky: Any reason why this is not an `enum class`?

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,118 @@
...
+        sh3_glvertarray(){Create();};
+
+        /**
+         *  VAO Attribute locations. These are used in a @ref sh3_glprogram to access data
+         *  sent to it from a VBO. These values were picked for no specific reason other than
+         *  the usual order of a vertex (vert data, normal data, uv data).
+         */
+        enum Attribute
+        {
+            SH3_VERTEX_SLOT = 0,
```
z33ky: I think automatic numbering should be fine here.

source/SH3/system/glbuffer.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "SH3/system/glbuffer.hpp"
+#include "SH3/system/log.hpp"
+
+#define GL_VBO_UNBIND 0
+
+using namespace sh3_gl;
+
+void sh3_glbuffer::Create()
+{
+    GLenum err;
```
z33ky: ?

source/SH3/system/glbuffer.cpp
```diff
@@ -0,0 +1,52 @@
...
+ *  @copyright 2017 Palm Studios
+ *
+ *  @date 8-6-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include "SH3/system/glbuffer.hpp"
+#include "SH3/system/log.hpp"
+
+#define GL_VBO_UNBIND 0
```
z33ky: Use `constexpr` instead.

source/SH3/system/glbuffer.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void sh3_glbuffer::Release()
+{
+    glDeleteBuffers(1, &id);
+}
+
+void sh3_glbuffer::Bind()
+{
+    GLenum err;
```
z33ky: ?

source/SH3/system/glprogram.cpp
```diff
@@ -48,6 +48,44 @@ std::string sh3_glprogram::load_error::message() const
     return error;
 }
 
+
+void sh3_glprogram::Load(const std::string& name)
```
z33ky: This should just call the other overload instead of duplicating the implementation.

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,72 @@
...
+ *  @copyright 2017 Palm Studios
+ *
+ *  @date 8-6-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include "SH3/system/glvertarray.hpp"
+#include "SH3/system/log.hpp"
+
+#define GL_VAO_UNBIND 0
```
z33ky: Use `constexpr`.

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,72 @@
...
+#define BUFFER_OFFSET(i) ((char *)NULL + (i))
```
z33ky: Use `constexpr`.

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+using namespace sh3_gl;
+
+void sh3_glvertarray::Create()
+{
+    GLenum err;
```
z33ky: ?

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    glGenVertexArrays(1, &id);
+
```
z33ky: Remove this line.

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void sh3_glvertarray::Release()
+{
+    glDeleteVertexArrays(1, &id);
+}
+
+void sh3_glvertarray::Bind()
+{
+    GLenum err;
```
z33ky: ?

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,72 @@
...
+    Bind();
+    glEnableVertexAttribArray(attrib); // Set attrib index as currently being used
+    buffers.at(attrib).Bind(); // Bind VBO we are going to send to
+    glVertexAttribPointer(attrib, size, type, GL_FALSE, stride, BUFFER_OFFSET(offset));
+    Unbind();
+}
+
+void sh3_glvertarray::AddVBO(sh3_glbuffer& buff)
+{
+    if(buffers.size() + 1 == 4) // If we have 4 VBOs, we know for sure we have an indexed VAO (Slot 1 VERTS, Slot 2 NORMALS, Slot 3 UV, Slot 4 INDEX)
```
z33ky: `if(buffers.size() == SH3_INDEX_SLOT)` makes the intent clear I think.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,118 @@
...
+         *  Draw our VertexArray
+         */
+        void Draw();
+
+
+    private:
+        GLuint                      id; /**< This VAO's ID. */
+        std::vector<sh3_glbuffer>   buffers; /**< All VBOs managed by this VAO. MUST BE IN ORDER SPECIFIED BY @ref Attribute. */
+
+        bool indexed = false;
```
z33ky: Replace with `bool is_indexed() const { return buffers.size() == *; }`?

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+        /**
+         *  Draw our VertexArray
+         */
+        void Draw();
+
+
+    private:
+        GLuint                      id; /**< This VAO's ID. */
+        std::vector<sh3_glbuffer>   buffers; /**< All VBOs managed by this VAO. MUST BE IN ORDER SPECIFIED BY @ref Attribute. */
```
z33ky: `sh3_glbuffer` seems small enough that we could just store an `std::array` of sh3_glbuffers with the maximum possible number of buffers, which is also a reasonably small 4 (this constant could also be included in the `Attribute` `enum` as `MAX` or something.
We then need an additional member to keep track of which buffers have been initialized. I guess an `std::underlying_type_t<Attribute>` which gets incremented when calling AddVBO does the job.

source/main.cpp
```diff
@@ -32,17 +47,45 @@ int main(int argc, char** argv)
...
+    sh3_gl::sh3_glbuffer color(sh3_gl::sh3_glbuffer::Target::SH3_ARRAY_BUFFER, "Color Data");
+
+    // VAO AND VBO TEST!
+    verts.BufferData(sizeof(g_vertex_buffer_data), g_vertex_buffer_data);
+    vao.AddVBO(verts);
+    vao.SetDataLocation(sh3_gl::sh3_glvertarray::Attribute::SH3_VERTEX_SLOT, sh3_gl::sh3_glvertarray::DataType::SH3_FLOAT, 3, 0, 0);
+
+    color.BufferData(sizeof(glCol), glCol);
+    vao.AddVBO(color);
+    vao.SetDataLocation(sh3_gl::sh3_glvertarray::Attribute::SH3_NORMAL_SLOT, sh3_gl::sh3_glvertarray::DataType::SH3_FLOAT, 3, 0, 0);
```
z33ky: If we pass the color as normal data, the API seems off.

z33ky (changes requested):
When changing the `enum`s to `enum class`es, then the `SH3_` prefix becomes unnecessary.

I'm not sold on the `AddVBO(); SetDataLocation();` API.
I think AddVBO should either return a handle, which can be passed to `SetDataLocation` instead of specifying the `Attribute`, or (preferably) a proxy-object, which you can call `SetDataLocation` on without having to specify any kind of `Attribute`.

Instead of using `vector.at(index)` I think we can just `assert(vector.size() >= index); vec[index]` to skip the bounds check in release-mode.

I'm not a huge fan of the way `main.cpp` is used to test whatever feature was most recently added.
It would make more sense to have each of this in a separate `main()` method in a separate file (and thus program).
However, this is not an issue with this pull request. We should open a new issue for this.

include/SH3/system/glbuffer.hpp
Quaker762: Yeah good idea, also considering this technically isn't a part of the actual game. I'll change it.

include/SH3/system/glbuffer.hpp
Quaker762: I'm not quite sure of the difference, could you possibly explain? All Google is telling me is that it's more strongly typed.

include/SH3/system/glbuffer.hpp
Quaker762: I'll get on that ASAP. There's no reason, in fact I should have done this in the reverse order.

include/SH3/system/glbuffer.hpp
Quaker762: I'm not quite sure how OpenGL works on this instance. Release frees the buffer's ID, and I'm not entirely sure if the data is released with it. If we create the buffers locally in say `sh3_model::Load( )`, and then release is called when the function returns, the data should be deleted from the GPU (and all the buffers in buffer becoming invalid)? I'll need to a bit more investigation, hence the commenting. I'll remove the line completely for now.

include/SH3/system/glbuffer.hpp
Quaker762: Okay I'll get on it. Need to study up a bit on templates.

source/SH3/system/glbuffer.cpp
Quaker762: Leftover code I should have removed. `glGetErorr( )` is actually a VERY bad function. It returns an error from any last glXYZ call without being cleared. We could have an issue in another part of the code that would trigger the error handler to be called, even if our buffer creation was successful. I'll remove these.

source/SH3/system/glvertarray.cpp
Quaker762: Remove `glGenVertexArrays(1, &id);`????

source/SH3/system/glvertarray.cpp
Quaker762: Wouldn't that imply that the next size would be 5, and the index buffer has already been added, hence crashing OpenGL?

EDIT: Nevermind, I forgot what section of code this was in.

source/main.cpp
Quaker762: This was just to test if the data was being passed correctly to the shader.

Quaker762:
> I think AddVBO should either return a handle, which can be passed to SetDataLocation

What handle would I return from AddVBO?? The whole point of it I suppose is to have a list of all current buffers loaded for a particular VAO. 

> instead of specifying the Attribute, or (preferably) a proxy-object, which you can call SetDataLocation on without having to specify any kind of Attribute.

How would I go about doing this? The only issue I can see is that `glVertexAttribPointer` requires an attribute in parameter one. I suppose I could add that to the buffer itself, instead of having an enum list. How would I pass a proxy object in??? IMHO it's easier to just have a list of the buffers and bind them, due to the way OpenGL works. Mainly I have it like this because it adheres to the way OpenGL requires us to do things without the annoyance of cluttering up the code with glXYZ calls everywhere and raw data manipulation. 

>I'm not a huge fan of the way main.cpp is used to test whatever feature was most recently added.
It would make more sense to have each of this in a separate main() method in a separate file (and thus program).

I probably should have just taken a screenshot, but I kind of wanted to demonstrate how the API for this works, as it was a little bit fiddly, and there are certain (calling) rules that need to be adhered to so that OpenGL doesn't shit the bed. 

Would it be better to have a whole lot of C functions then, rather than implementing this in classes??

source/SH3/system/glprogram.cpp
Quaker762: I think OpenGL will crash if I specify no attributes, I'll have to double check. I need to research a bit more to better abstract some of this stuff.

include/SH3/system/glbuffer.hpp
Quaker762: > "by the C++ Runtime"??

Doesn't the runtime library handle all this? Regardless I'll remove it.

include/SH3/system/glbuffer.hpp
z33ky: Yes, that is one difference. This means there's no implicit cast to or from numeric types (`int x = SH3_ARRAY_BUFFER; Target y = 5;`). This is still possible, it just needs explicit casts.
And scoping is enforced, so the `Target::` prefix is required.

include/SH3/system/glbuffer.hpp
z33ky: That'd be true if we'd be using a garbage collected language.
Perhaps you could say that using a smart pointer would also qualify, since the standard library includes the code to invoke the destructor.
Usually however, destructor calls are inserted at the end of scopes by the compiler, requiring no C++ runtime to even be present.

source/SH3/system/glvertarray.cpp
z33ky: No, the empty line after that :)

source/SH3/system/glprogram.cpp
z33ky: Is this method not equivalent to the other Load-overload when called with an empty `attribs` vector?

source/SH3/system/glbuffer.cpp
z33ky: We can define a `SH3ClearGLError` function, which pops all errors of the error queue and warns about them. But this should be in a separate commit. Even in a separate pull request if you want.

include/SH3/system/glbuffer.hpp
z33ky: Seems to work, if we want to trust [this](https://stackoverflow.com/a/19645105) StackOverflow answer.

z33ky:
> What handle would I return from AddVBO?? The whole point of it I suppose is to have a list of all current buffers loaded for a particular VAO.

And you would keep that. AddVBO would just return a handle, like the index of the most recently added object, to identify the VBO that was added. This is used by `SetDataLocation` to identify the object.
In fact, the index already exists, its value just assumed to be whatever `enum Attribute` constant.

> How would I go about doing this? The only issue I can see is that `glVertexAttribPointer` requires an attribute in parameter one. I suppose I could add that to the buffer itself, instead of having an enum list. How would I pass a proxy object in??? IMHO it's easier to just have a list of the buffers and bind them, due to the way OpenGL works. Mainly I have it like this because it adheres to the way OpenGL requires us to do things without the annoyance of cluttering up the code with glXYZ calls everywhere and raw data manipulation.

Something like this:
```
struct ProxyObject final
{
    Attribute self;
    sh3_glvertarray &array;
    void SetDataLocation(...) { array.SetDataLocation(self, ...); }
};
```
I think you misunderstood what I had in mind with a proxy object.

> I probably should have just taken a screenshot, but I kind of wanted to demonstrate how the API for this works, as it was a little bit fiddly, and there are certain (calling) rules that need to be adhered to so that OpenGL doesn't shit the bed.

It makes sense, both to illustrate the API like you stated, and also to have a test ready in case someone wants to change something and test that it still works.
I don't think removing/commenting out whatever was last in `main()` is a good approach though, since the next thing to test is going to come along and `main()` will be cluttered up some more (or if the code is removed, the example/test case is gone).

> Would it be better to have a whole lot of C functions then, rather than implementing this in classes??

I didn't state that the OO encapsulation was bad. In contrast, I think it's a good approach. I just don't like the implicit assumptions about the `sh3_vertarray::buffers`-index (i.e. the values of `enum Attribute` to use).

source/SH3/system/glprogram.cpp
Quaker762: It is. It's been a while since I've looked over the `glprogram` code, so I kinda forgot haha.

Quaker762:
> And you would keep that. AddVBO would just return a handle, like the index of the most recently added object, to identify the VBO that was added. This is used by SetDataLocation to identify the object.
In fact, the index already exists, its value just assumed to be whatever enum Attribute constant.

Ohh I see, that's a much better way of handling it, I didn't think of that.

> I think you misunderstood what I had in mind with a proxy object.

So basically something that encapsulates the VAO without having to pass an attribute? I'm not quite sure that will work with what I have planned for the 3D model section.

> I don't think removing/commenting out whatever was last in main() is a good approach though, since the next thing to test is going to come along and main() will be cluttered up some more (or if the code is removed, the example/test case is gone).

I see, I have the same sentiment. Are there any testing frameworks we can use (like jUnit), or is better to have multiple mains in a `test` folder and then have test targets?? Might come in handy for a 3D camera test.

> I just don't like the implicit assumptions about the sh3_vertarray::buffers-index (i.e. the values of enum Attribute to use).

My reasoning is that (at least from my deductions), that it's up the programmer to decide which attributes are bound where. For example, we could have it in the order: UV, vertex, normals, but that's a pretty bad idea. It's also not _really_ an implicit assumption per se, as SILENT HILL's vertex format is in this **exact** same order, so it's technically consistent. But I do agree, the handle idea does make it a lot cleaner and makes a lot more sense, though an attribute ID is technically different to an attribute index. I've gone and changed it to the handle idea, but I still think having a weakly typed regular `enum` to pass to OpenGL is a good way to handle it.

source/SH3/system/glbuffer.cpp
Quaker762: I'll get on that ASAP.

include/SH3/system/glprogram.hpp
Quaker762: Fixed, I had a bit of brainfart in my rush to implement this.

include/SH3/system/glbuffer.hpp
Quaker762: Interesting. The information for this is so fragmented, thanks for finding this haha.

Quaker762:
I've just stumbled upon another GL abstraction library here: https://github.com/Overv/OOGL/blob/150aebfa307a76119f62085cb0d73991793c0dbe/include/GL/GL/VertexArray.hpp

It seems that this author came to almost the same conclusion I did for his VAOs and VBOs (I'm taking a look at other C++ abstractions of OpenGL data objects)

z33ky:
> So basically something that encapsulates the VAO without having to pass an attribute? I'm not quite sure that will work with what I have planned for the 3D model section.

`Attribute a = vao.AddVBO(o); vao.SetDataLocation(a, ...);` is equivalent to `Proxy p = vao.AddVBO(o); p.SetDataLocation(...)`.

> I see, I have the same sentiment. Are there any testing frameworks we can use (like jUnit), or is better to have multiple mains in a `test` folder and then have test targets??

There is no one big C++ testing framework. A colleague of mine recently reviewed testing frameworks of various open source C libraries, I'll ask him if he knows some good C++ ones.

To start with, I think just a test-directory with one target (executable) per test and no automatic testing is fine.

> My reasoning is that (at least from my deductions), that it's up the programmer to decide which attributes are bound where. For example, we could have it in the order: UV, vertex, normals, but that's a pretty bad idea. It's also not _really_ an implicit assumption per se, as SILENT HILL's vertex format is in this **exact** same order, so it's technically consistent.

The assumption being backed by the data layout does not make it explicit.
In the current version, the attribute that gets modified by a call to `SetDataLocation` is not visible in the call itself, it can only be implicitly from the context where (and when) it is called.

> But I do agree, the handle idea does make it a lot cleaner and makes a lot more sense, though an attribute ID is technically different to an attribute index.

Right. If I understand this right, an attribute ID is a higher level concept represented by the `enum Attribute`, right? While the attribute index is the index into `sh3_glvertarray::buffers`?
And it "just happens" that there's a 1-to-1 mapping from ID to index, yeah?

Anyways, if this change is adopted then the `enum Attribute` is not even needed, at least not in this class. It can instead be moved to the `class texture`.
We might also want to use this VAO abstraction for procedurally generated objects, e.g. quads, boxes or spheres used for debugging, where we might only have vertices (SH3_VERTEX_SLOT) and indices (SH3_INDEX_SLOT), but no texture. So there we would use a different enumeration.

Quaker762:
> Attribute a = vao.AddVBO(o); vao.SetDataLocation(a, ...); is equivalent to Proxy p = vao.AddVBO(o); p.SetDataLocation(...).

Ohhh okay I get it now.

> To start with, I think just a test-directory with one target (executable) per test and no automatic testing is fine.

Yeah this sounds much better. Would it perhaps be feasible to compile the engine as libsh3.a or something like that (basically an engine/game detachment). I'll open an issue if you think this might work.

> The assumption being backed by the data layout does not make it explicit.
In the current version, the attribute that gets modified by a call to SetDataLocation is not visible in the call itself, it can only be implicitly from the context where (and when) it is called.

I think I see what you mean now, so the caller knows what type was passed in, but the actual _function_ itself doesn't know what type was explicitly passed (because of the weak typing)?

> Right. If I understand this right, an attribute ID is a higher level concept represented by the enum Attribute, right? While the attribute index is the index into sh3_glvertarray::buffers?
And it "just happens" that there's a 1-to-1 mapping from ID to index, yeah?

Yeah, it's sort of hard to explain. At the end of the day, it's all got to do with _how_ the streams are passed into the vertex shader so we can perform operations on them (using the `layout(location = 0) in` style syntax). It makes more sense to hardcode the positions, because, as you've said, we might bind a quad with only vertex index and uv data, which makes our shaders consistent as well.

> Anyways, if this change is adopted then the enum Attribute is not even needed, at least not in this class. It can instead be moved to the class texture.

How so? The attributes are part of the VAO (even part of the glprogram is you can spin it like that I suppose haha). 

> We might also want to use this VAO abstraction for procedurally generated objects, e.g. quads, boxes or spheres used for debugging, where we might only have vertices (SH3_VERTEX_SLOT) and indices (SH3_INDEX_SLOT), but no texture. So there we would use a different enumeration.

Yes!! This is point of abstracting it the way I did. Not all geometry will be laid out the same way, and having (if you will) "pre-baked" locations makes it _much_ simpler imho when it comes to writing a shader, as we know what to expect in which location. Although, I do see what you mean by having the attributes per class (like having them in texture so we can map vertex to 0, uv to 1 and index to 2 etc, which will work given the current implementation).

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,109 @@
...
+        {
+            ARRAY_BUFFER    = GL_ARRAY_BUFFER,
+            ELEMENT_BUFFER  = GL_ELEMENT_ARRAY_BUFFER,
+            TEXTURE_BUFFER  = GL_TEXTURE_BUFFER
+        };
+
+        /**
+         *  Default constructor.
+         */
+        sh3_buffer(Target type, const std::string& buffname = ""){buffType = type, name = buffname;Create();};
```
z33ky: Use initializer lists (`(... buffname = ""):bufftype(type), name(buffname)`).

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+        /**
+         *  Destructor. Ensure we have no leaks when this buffer is to be deleted by the C++ Runtime.
+         */
+        ~sh3_buffer(){/**Release();*/};
```
z33ky: This should be fine, right? Otherwise put a `//FIXME` there explaining why this is not done currently.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,109 @@
...
+         */
+        void Release();
+
+        /**
+         *  Give some data to our buffer
+         *
+         *  @param num Size (in bytes) of the data we are flushing.
+         *  @param data void pointer to our data (usually an std::array or std::vector).
+         */
+        void BufferData(void* data, std::size_t num);
```
z33ky: `sh3_arc_file::ReadObject`-like interface? You should be able to just copy it over and adapt the parameters, even if you're not up to speed on templates.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,109 @@
...
+         *  Get the number of elements contained within this buffer.
+         *
+         *  @return Number of elements in this GL Buffer.
+         */
+        std::uint32_t GetNumberElements() const;
+
+    private:
+        GLuint          id;         /**< The ID of our buffer given to us by OpenGL */
+        Target          buffType;   /**< What type of buffer this is */
+        std::uint32_t   size;      /**< The size of this buffer in bytes */
```
z33ky: Misaligned comment :S

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,126 @@
...
+        vertarray(){Create();};
+
+        /**
+         *  VAO Attribute locations. These are used in a @ref program to access data
+         *  sent to it from a VBO. These values were picked for no specific reason other than
+         *  the usual order of a vertex (vert data, normal data, uv data).
+         */
+        enum class Attribute
+        {
+            VERTEX_SLOT = 0,
```
z33ky: `enum`s start with 0 if you don't specify it.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,126 @@
...
+        std::vector<sh3_buffer>   buffers; /**< All VBOs managed by this VAO. MUST BE IN ORDER SPECIFIED BY @ref Attribute. */
+
+        /**
+         *  Determine if this VAO is using indexed drawing or not
+         *
+         *  @return bool - Does this VAO contain an index buffer?
+         */
+        bool is_indexed() const
+        {
+            return buffers.size() == 4;
```
z33ky: Should be `== static_cast<std::underlying_type_t<Attribute>(INDEX_SLOT)`.

z33ky (changes requested):
You skipped many of the requested improvements :(
Why is that?

Quaker762:
> You skipped many of the requested improvements :(
Why is that?

Aside from what I emailed you about, I _maaaay_ have forgotten to do a git stash pop at 12am sorry HAHA.

include/SH3/system/glbuffer.hpp
Quaker762: I fixed this, stashed it and then forgot to unstash.. 

Sorry haha.

include/SH3/system/glvertarray.hpp
Quaker762: Ahhh good I wasn't entirely sure if it didn't without specifying at least one.

Quaker762:
@z33ky 

I've changed the way this works to your 'handle' idea. It actually works much better than how I originally had it, and also makes sure correct order is maintained across the VAO and shader. It should work really well, but until we have some real OpenGL up and running, it's hard to tell haha.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+        /**
+         *  Default constructor.
+         */
+        sh3_buffer(Target type, const std::string& buffName = ""): buffType(type), name(buffName){Create();};
+
+        /**
+         *  Destructor. Ensure we have no leaks when this buffer is to be deleted by the C++ Runtime.
+         */
+        ~sh3_buffer(){/**Release(); FIXME: This causes a segmentation violation (why is this called in the game loop!?!?!)*/};
```
Quaker762: @z33ky This causes a segfault in the colour buffer code in main.cpp. It is seemingly cleared from the time it is created to the time it is needed in the game loop.. Could the compiler possibly be auto optimising this in???

Quaker762:
@z33ky Uh-oh....

Aside from the conflict (which I thought would happen...), it's good to go! No more 2982839428374 number of elements to draw (and grinding the program to a halt haha).

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+        {
+            ARRAY_BUFFER    = GL_ARRAY_BUFFER,
+            ELEMENT_BUFFER  = GL_ELEMENT_ARRAY_BUFFER,
+            TEXTURE_BUFFER  = GL_TEXTURE_BUFFER
+        };
+
+        /**
+         *  Default constructor.
+         */
+        buffer_object(Target type = Target::ARRAY_BUFFER, const std::string& buffName = ""): buffType(type), name(buffName){Create();};
```
z33ky: Slap a FIXME here, since we only set a default value for `type`, because we're not passing anything when creating the buffers for the VAO.
Look at my mail from the 17th for some suggestions. I like the last one the most, where we pair the SLOT-enum and the targets in a struct.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+         *  Free any data related to this buffer, instead of letting the destructor do it.
+         *  We explicitly tell OpenGL to release the buffer.
+         */
+        void Release();
+
+        /**
+         *  Give some data to our buffer
+         *
+         *  @param num Size (in bytes) of the data we are flushing.
+         *  @param data void pointer to our data (usually an std::array or std::vector).
```
z33ky: `usage` docs are missing.
Reorder `data` before `num`.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+         */
+        void BufferData(void* data, std::size_t num, GLenum usage);
+
+        /**
+         *  Redefines/updates a subset of data stored in this GLBuffer.
```
z33ky, line 73: Parameter docs are missing.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+            TEXTURE_BUFFER  = GL_TEXTURE_BUFFER
+        };
+
+        /**
+         *  Default constructor.
+         */
+        buffer_object(Target type = Target::ARRAY_BUFFER, const std::string& buffName = ""): buffType(type), name(buffName){Create();};
+
+        /**
+         *  Destructor. Ensure we have no leaks when this buffer is to be deleted by the C++ Runtime.
```
z33ky: See https://github.com/Palm-Studios/sh3redux/pull/94#discussion_r122571809.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+         *  Realease our VAO.
+         */
+        void Release();
+
+        /**
+         *  Bind this VAO.
+         */
+        void Bind();
+
+         /**
```
z33ky: Incorrect indent.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+          * Unind this VAO.
+          */
+        void Unbind();
+
+          /**
```
z33ky: Incorrect indent.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+        std::array<buffer_object, numbuffers> buffers; /**< Our VBOs. */
+
+    private:
+
+        /**
+         *  Get the offset (stride) of a particular variable (e.g colors mixed with vertices) in a data buffer.
+         *
+         *  @param  offset - Offset into the buffer
+         *
+         *  @return void* - Pointer to value.
```
z33ky: Type needs not be specified in doc.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+         */
+        static constexpr void* buffer_offset(GLint offset){return static_cast<char *>(NULL) + offset;}
```
z33ky: Use `nullptr` instead of `NULL`. The usage here is fine, but in other places `NULL` can select incorrect overloads since it's an integer in C++. So we should just generally avoid it.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+    };
+
+    template<std::size_t numbuffers>
+    void mutablevao<numbuffers>::SetDataLocation(GLuint attrib, DataType type, GLint size, GLsizei stride, GLint offset)
+    {
+        Bind();
+        glEnableVertexAttribArray(attrib); // Set attrib index as currently being used
+        ASSERT_MSG(attrib < numbuffers, "VBO index (attribute) out of range");
+        buffers[attrib].Bind(); // Bind VBO we are going to send to
```
z33ky: The code seems to clearly express what's being done without the comment.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+         *  @return void* - Pointer to value.
+         */
+        static constexpr void* buffer_offset(GLint offset){return static_cast<char *>(NULL) + offset;}
+    };
+
+    template<std::size_t numbuffers>
+    void mutablevao<numbuffers>::SetDataLocation(GLuint attrib, DataType type, GLint size, GLsizei stride, GLint offset)
+    {
+        Bind();
+        glEnableVertexAttribArray(attrib); // Set attrib index as currently being used
```
z33ky: The code seems to clearly express what's being done without the comment.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,311 @@
...
+        {
+            ASSERT_MSG(attrib < Attribute::MAX, "VBO index (attribute) out of range");
+            vaoparent<Attribute>::type::SetDataLocation(static_cast<GLuint>(attrib), type, size, stride, offset);
+        }
+    };
+
+    /**
+     *  A wrapper for an OpenGL 3.3 Vertex Array Object.
+     */
+    struct vertarray final
```
z33ky: This can be removed now, right?

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,94 @@
...
+void finalvao::Draw()
+{
+    Bind();
+    if(ibo)
+        glDrawElements(GL_TRIANGLES, ibo->GetNumberElements(), GL_UNSIGNED_INT, 0);
+    else
+        glDrawArrays(GL_TRIANGLES, 0, vertices.GetNumberElements());
+}
+
+//void vertarray::Create()
```
z33ky: Remove?

source/SH3/system/glvertarray.cpp
```diff
@@ -0,0 +1,94 @@
...
+ *
+ *  @date 8-6-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include "SH3/system/glvertarray.hpp"
+#include "SH3/system/log.hpp"
+
+constexpr int   GL_VAO_UNBIND = 0;
+constexpr void* BUFFER_OFFSET(std::uint32_t i){return((char *)NULL + (i));}
```
z33ky: remove

source/main.cpp
```diff
@@ -32,17 +54,46 @@ int main(int argc, char** argv)
...
-    sh3_graphics::msbmp("data/pic/error.bmp");
+    bool quit = false;
+    SDL_Event e;
+
+    sh3_gl::program::load_error err;
+    sh3_gl::program test("test", err);
+    sh3_gl::vao<MyAttributes> myVao;
+
+    myVao.Bind();
+    sh3_gl::buffer_object verts = myVao[MyAttributes::VERTEX_SLOT];
```
z33ky: This can be a reference.
In fact I'm not sure if there's a good reason to have `buffer_object` and `vao` being copyable. We may just want to delete their copy ctor. If we find a use-case for it later, we can still change them to allow it.

z33ky (changes requested):
some `*vao` stuff is missing documentation.

The documentation for `vao::Attribute` should mention our assumptions, e.g. there's a `Attribute::MAX` and `Attribute::VERTEX_SLOT` and VAO using an IBO have an `Attribute::INDEX_SLOT`.
Note that Doxygen has a keyword to document template parameters: `tparam`.

For some parameters, you use `@param name - doc` - I'm not sure if Doxygen removes the dash.

It seems to me that the `glEnableVertexAttribArray` is missing a symmetric `glDisableVertexAttribArray` call, or is this not needed?

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,126 @@
...
+            ARRAY_BUFFER    = GL_ARRAY_BUFFER,
+            ELEMENT_BUFFER  = GL_ELEMENT_ARRAY_BUFFER,
+            TEXTURE_BUFFER  = GL_TEXTURE_BUFFER
+        };
+
+        /**
+         *  Constructor.
+         *
+         *  FIXME: We're currently <i>always</i> passing @ref Target::ARRAY_BUFFER to the buffer object,
+         *  as we never actually pass anything to this constructor.
```
z33ky: Note that we don't pass anything in `@ref mutablevao`.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,126 @@
...
+         */
+        buffer_object(Target type = Target::ARRAY_BUFFER, const std::string& buffName = ""): buffType(type), name(buffName){Create();};
+
+        /**
+         *  Destructor. Ensure we have no leaks when this buffer is to be deleted.
```
z33ky: Destructor implies that memory gets freed. I think just "Destructor." suffices.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,126 @@
...
+         *  We explicitly tell OpenGL to release the buffer.
+         */
+        void Release();
+
+        /**
+         *  Give some data to our buffer
+         *
+         *  @param data void pointer to our data (usually an std::array or std::vector).
+         *  @param num Size (in bytes) of the data we are flushing.
+         *  @param usage - The way in which want this data stored (i.e, can we rewrite sections of it?).
```
z33ky: Remove the dash.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,126 @@
...
+        /**
+         *  Free any data related to this buffer, instead of letting the destructor do it.
+         *  We explicitly tell OpenGL to release the buffer.
+         */
+        void Release();
+
+        /**
+         *  Give some data to our buffer
+         *
+         *  @param data void pointer to our data (usually an std::array or std::vector).
```
z33ky: I'd just say "pointer to our data".

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,126 @@
...
+         *  @param num Size (in bytes) of the data we are flushing.
+         *  @param usage - The way in which want this data stored (i.e, can we rewrite sections of it?).
+         */
+        void BufferData(void* data, std::size_t num, GLenum usage);
+
+        /**
+         *  Redefines/updates a subset of data stored in this GLBuffer.
+         *
+         *  @param data void pointer to our data (usually an std::array or std::vector).
```
z33ky: Same as for BufferData.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+#include "SH3/system/assert.hpp"
+#include "SH3/system/glbuffer.hpp"
+
+namespace sh3_gl{
+
+    /**
+     *  Contains actual functionality of our VAO (@ref Bind, @ref Create etc).
+     *
+     *  @note Cannot be directly instantiated, use @ref vao instead
```
z33ky: Can != should. This should not be directly instantiated.
Also, add a trailing dot.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,249 @@
...
+    {
+    private:
+        template<typename T>
+        static constexpr indexedmutablevao<T::MAX, T::INDEX_SLOT> test(decltype(T::INDEX_SLOT)*);
+
+        template<typename T>
+        static constexpr mutablevao<T::MAX> test(...);
+
+    public:
+        using type = decltype(test<Attribute>(0)); /**< Type of vao this is (either @ref mutablevao or @ref indexedmutablevao)
```
z33ky: Missing `*/`.

tests/vao.cpp
```diff
@@ -0,0 +1,100 @@
...
+};
+
+enum MyAttributes
+{
+    VERTEX_SLOT,
+    COLOR_SLOT,
+    MAX
+};
+
+/**
```
z33ky: I don't think we need to Doxygen-comment our tests.
If we write our own testing framework, then that should be documented, but that's still in the future, if ever.

tests/vao.cpp
```diff
@@ -0,0 +1,100 @@
...
+    SDL_Event e;
+
+    sh3_gl::program::load_error err;
+    sh3_gl::program test("test", err);
+    sh3_gl::vao<MyAttributes> myVao;
+
+    myVao.Bind(); // Bind our VAO
+	
+	// Create the vertex buffer object
+    sh3_gl::buffer_object verts = myVao[MyAttributes::VERTEX_SLOT];
```
z33ky: Reference, as in main.cpp.

tests/vao.cpp
```diff
@@ -0,0 +1,100 @@
...
+    verts.BufferData(g_vertex_buffer_data, sizeof(g_vertex_buffer_data), GL_STATIC_DRAW);
+    myVao.SetDataLocation(MyAttributes::VERTEX_SLOT, sh3_gl::vao<MyAttributes>::DataType::FLOAT, 3, 0, 0);
+
+	// Create the color buffer object
+    sh3_gl::buffer_object col = myVao[MyAttributes::COLOR_SLOT];
```
z33ky: Reference, as in main.cpp.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,249 @@
...
+            Create();
+        };
+
+        //No virtual dtor, because finalvao should not be stored directly, i.e. this should only ever be used as a reference.
+
+    private:
+        //Prevent dynamic instantiation since we have no virtual dtor.
+        static void* operator new(std::size_t) = delete;
+        static void* operator new[](std::size_t) = delete;
+
```
z33ky: Also delete the 'delete` operators.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,249 @@
...
+         */
+        finalvao(const buffer_object& verts):ibo(nullptr), vertices(verts)
+        {
+            Create();
+        };
+
+        /**
+         *  Indexed constructor.
+         */
+        finalvao(const buffer_object &ibo, GLsizei vertex_count, const buffer_object& verts):ibo(&ibo), vertices(verts)
```
z33ky: Remove `vertex_count`.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,249 @@
...
+        GLuint GetID() const {return id;}
+
+        /**
+         *  Draw our VAO.
+         */
+        void Draw();
+
+    protected:
+        GLuint  id;                         /**< Our VAO's identification number. **/
+        GLsizei vertex_count;               /**< The number of vertices to draw. **/
```
z33ky: Remove.

SILENT HILL 3 Redux.cbp
```diff
@@ -134,10 +138,14 @@
...
 		<Unit filename="source/SH3/arc/vfile.cpp" />
 		<Unit filename="source/SH3/graphics/msbmp.cpp" />
+		<Unit filename="source/SH3/graphics/quad.cpp" />
 		<Unit filename="source/SH3/graphics/texture.cpp" />
+		<Unit filename="source/SH3/system/assert.cpp" />
 		<Unit filename="source/SH3/system/config.cpp" />
+		<Unit filename="source/SH3/system/glbuffer.cpp" />
 		<Unit filename="source/SH3/system/glcontext.cpp" />
 		<Unit filename="source/SH3/system/glprogram.cpp" />
+		<Unit filename="source/SH3/system/glvertarray.cpp" />
```
z33ky, line 148: Add those to `source/CMakeLists.txt` as well.

z33ky (changes requested):
Some *vao ctors are missing parameter or any documentation.
There's still some dashes following the parameter names in `@param` docs.

Sorry, this drags on so long, but after these superficial things are fixed I'll approve it. I promise ;)

Quaker762:
> Sorry, this drags on so long, but after these superficial things are fixed I'll approve it. I promise ;)

Haha it's no issue, I'd rather have the code nice as well instead of being half arsed.

SILENT HILL 3 Redux.cbp
Quaker762: They're already in there, hence the merge conflict.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,127 @@
...
+         *  We explicitly tell OpenGL to release the buffer.
+         */
+        void Release();
+
+        /**
+         *  Give some data to our buffer
+         *
+         *  @param data void pointer to our data.
+         *  @param num Size (in bytes) of the data we are flushing.
+         *  @param usage - The way in which want this data stored (i.e, can we rewrite sections of it?).
```
z33ky: Remove the dash here.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+     *  @note Should <i>not</i> be directly instantiated, use @ref vao instead.
+     */
+    struct finalvao
+    {
+    protected:
+
+        /**
+         *  Non indexed constructor.
+         */
+        finalvao(const buffer_object& verts):ibo(nullptr), vertices(verts)
```
z33ky: @Param docs missing.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+        {
+            Create();
+        };
+
+        /**
+         *  Indexed constructor.
+         */
+        finalvao(const buffer_object &ibo, const buffer_object& verts):ibo(&ibo), vertices(verts)
```
z33ky: Param docs missing.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+            USHORT  = GL_UNSIGNED_SHORT,
+            INT     = GL_INT,
+            UINT    = GL_UNSIGNED_INT,
+            FLOAT   = GL_FLOAT
+        };
+
+        /**
+         *  Set vertex shader location of data in VBO for specified attribute.
+         *
+         *  @param attrib - Attribute we want to map data input in our shader (using location = x).
```
z33ky: Remove the dashes here.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+         *  @param size - The size of each component per attribute.
+         *  @param stride - Byte offset between vertex attributes.
+         *  @param offset - Offset to attribute in buffer. Each attrib is then located at offset + stride.
+         */
+        void SetDataLocation(GLuint attrib, DataType type, GLint size, GLsizei stride, GLint offset);
+
+        /**
+         *  Overload the [] operator to get a buffer stored in @ref buffers more easily.
+         *
+         *  @param slot - Index slot where the buffer is stored. Usually in a @ref Attribute.
```
z33ky: Remove the dash here.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+    protected:
+        std::array<buffer_object, numbuffers> buffers; /**< Our list of current VBOs. */
+
+    private:
+
+        /**
+         *  Get offset (stride) into a buffer stored in @ref buffers
+         *
+         *  @param offset - Offset to the first element.
```
z33ky: Remove the dash here.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+         *
+         *  @return void* - Pointer to the first element of stride 'offset'.
```
z33ky: And here.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+         */
+        static constexpr void* buffer_offset(GLint offset){return static_cast<char *>(nullptr) + offset;}
+    };
+
+    /**
+     *  Set vertex shader location of data in VBO for specified attribute.
+     *
+     *  @tparam numbuffers - Number of buffers
```
z33ky: Remove dashes here.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+    /**
+     *  A VAO with indexed vertices.
+     *
+     *  You probably want to use @ref vao.
+     */
+    template<std::size_t numbuffers, std::size_t index_slot>
+    struct indexedmutablevao : public mutablevao<numbuffers>
+    {
+    protected:
+        indexedmutablevao(const buffer_object& vertices): mutablevao<numbuffers>(mutablevao<numbuffers>::buffers[index_slot], numbuffers, vertices) {}
```
z33ky: Documentation missing here. Don't forget the parameter ;)

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+    };
+
+    /**
+     *  Declares parent type for @ref vao.
+     *
+     *  Either @ref indexedmutablevao or @ref mutablevao (unindexed).
+     *
+     *  @tparam Attribute - Attribute enum (specifies Attribute::MAX to send to @ref mutablevao).
```
z33ky: Remove the dash here.

z33ky: Also document the usage of `Attribute::INDEX_SLOT`.

I'd just leave a short parameter description for `@tparam` and put the description of it below:
```
@tparam Attribute An attribute enum.

@p Attribute should have a value @c MAX to indicate the number of attributes (slots).
If @p Attribute has a value @c INDEX_SLOT, then the @ref type is an @ref indexedmutablevao.
```

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+    public:
+        using type = decltype(test<Attribute>(0)); /**< Type of vao this is (either @ref mutablevao or @ref indexedmutablevao) */
+
+    private:
+        vaoparent() = delete;
+    };
+
+    template<typename Attribute>
+    struct vao final : public vaoparent<Attribute>::type
```
z33ky: Missing docs.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+        /**
+         *  Constructor.
+         */
+        vao(): vaoparent<Attribute>::type(vaoparent<Attribute>::type::buffers[Attribute::VERTEX_SLOT]){}
+
+        /**
+         *  Set vertex shader location of data in VBO for specified attribute.
+         *
+         *  @param attrib - Attribute we want to map data input in our shader (using location = x).
```
z33ky: Remove the dashes.

z33ky (changes requested):
Did you test whether Doxygen ignores the dashes after parameter names?
I now marked all places where this occurs.

In my code I aligned the parameter documentation, which also visually stands out, which I assume is why you used the dashes.
```
@param data  Void pointer to our data.
@param num   Size (in bytes) of the data we are flushing.
@param usage The way in which want this data stored (i.e, can we rewrite sections of it?).
```

I also pointed out where exactly documentation is missing.

z33ky:
Also, to resolve the merge conflicts you should rebase your code on master.

Quaker762:
> Did you test whether Doxygen ignores the dashes after parameter names?

I just checked, it keeps them in there. I'm just used to using dashes in a list of things, but aligning is a lot cleaner. I'll remove all of them.

> I also pointed out where exactly documentation is missing.

Whoops, I must have missed some of that sorry.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+     *  A VAO with indexed vertices.
+     *
+     *  You probably want to use @ref vao.
+     *
+     *  @tparam numbuffers  Number of buffers that should be reserved in @ref buffers.
+     *  @tparam index_slot  The slot number where our index @ref buffer_object resides.
+     */
+    template<std::size_t numbuffers, std::size_t index_slot>
+
+    /**
```
z33ky: This should be moved to the ctor.
Also, VAo -> VAO.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+     *  VAo that inherits from @ref mutablevao. This passes our index @ref buffer_object to the @ref finalvao cosntructor.
```
z33ky: Too implementation specific. Just "Constructor" is fine.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+     *
+     *  @param vertices Reference to our vertices @ref buffer_object. This gets passed to @ref finalvao.
```
z33ky: " This gets passed to @ref finalvao.": unnecessary implementation detail.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+    protected:
+        indexedmutablevao(const buffer_object& vertices): mutablevao<numbuffers>(mutablevao<numbuffers>::buffers[index_slot], numbuffers, vertices) {}
+    };
+
+    /**
+     *  Declares parent type for @ref vao.
+     *
+     *  Either @ref indexedmutablevao or @ref mutablevao (unindexed).
+     *
+     *  @tparam Attribute Attribute enum (specifies Attribute::MAX to send to @ref mutablevao).
```
z33ky: See https://github.com/Palm-Studios/sh3redux/pull/94#discussion_r128913919.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+    private:
+        vaoparent() = delete;
+    };
+
+    /**
+     *  Our VAO object. This handles all @ref buffer_objects needed to draw data that has been
+     *  sent to the graphics card.
+     *
+     *  @tparam Attribute The attribute enum that contains all slots required by this VAO. This
```
z33ky: This should also mention `VERTEX_SLOT`, `MAX` and optional `INDEX_SLOT`.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+     *  makes it more generic, i.e, some VAOs might need a color buffer instead of a UV buffer etc.
```
z33ky: I'd indent it so it looks like
```
     *  @tparam Attribute The attribute enum that ...
     *                    makes it more generic, ...
```

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,287 @@
...
+     */
+    template<std::size_t numbuffers>
+    struct mutablevao : public finalvao
+    {
+    protected:
+
+        /**
+         *  Our VAO that inherits from @ref finalvao. Contains a list of all our @ref buffer_object structures.
+         *
+         *  @param vertices Reference to our vertices @ref buffer_object. This gets passed
```
z33ky: " This gets passed to @ref": this sentence is missing an end. Also, unnecessary implementation detail. Just omit it.

z33ky (changes requested):

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     *  A VAO with indexed vertices.
+     *
+     *  You probably want to use @ref vao.
+     *
+     *  @tparam numbuffers  Number of buffers that should be reserved in @ref buffers.
+     *  @tparam index_slot  The slot number where our index @ref buffer_object resides.
+     */
+    template<std::size_t numbuffers, std::size_t index_slot>
+
```
z33ky: Remove this empty line.

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,291 @@
...
+    private:
+        vaoparent() = delete;
+    };
+
+    /**
+     *  Our VAO object. This handles all @ref buffer_objects needed to draw data that has been
+     *  sent to the graphics card.
+     *
+     *  @tparam Attribute   The attribute enum that contains all slots required by this VAO. This
```
z33ky: This is a very long parameter description. Do something like the `Attribute` doc for `vaoparent`.

z33ky (changes requested):

z33ky:
Ah, and don't forget to use `@c` and `@p` like for the `Attribute` of `vaoparent`. It instructs Doxygen to use typewriter font.

Quaker762:
> Ah, and don't forget to use @c and @p like for the Attribute of vaoparent.

Where did you want me to put that?? `@tparam @p Attribute`???

z33ky:
No. See https://github.com/Quaker762/sh3redux/blob/8071b83685bfcaaa0686e5deb076b249edbf8c1f/include/SH3/system/glvertarray.hpp#L232. Use `@c` to refer to values (including symbolic ones like enum members) and `@p` to refer to the parameter (outside of `@param Attribute`).
For forget what the difference between the two commandos is, but it made sense this way when I looked them up.

Quaker762:
@z33ky This should be all good to go unless I've missed anything else (which I hope I haven't haha)

include/SH3/system/glvertarray.hpp
```diff
@@ -0,0 +1,288 @@
...
+
+    private:
+        vaoparent() = delete;
+    };
+
+    /**
+     *  Our VAO object. This handles all @ref buffer_objects needed to draw data that has been.
+     *  sent to the graphics card.
+     *
+     *  @tparam Attribute   @p Attribute enum that contains all slots required by this VAO, making it
```
z33ky: This is still a very long parameter description.
The idea is to be brief here and go into detail below.
Look how `Attribute` is documented for `vaoparent`.

z33ky (approved):

Quaker762:
Now we're rolling ;)

PR #97, opened 2017-07-09
Tweaked GL Program interface

Quaker762:
Now allows the programID to be accessed by the caller. Useful form interfacing with a GLSL program from C++. Also adds an interface to find uniform locations from the currently active GL Program.

### DIFF
--- a/include/SH3/system/glprogram.hpp
+++ b/include/SH3/system/glprogram.hpp
@@ -17,15 +17,13 @@
 
 #include <GL/glew.h>
 #include <GL/gl.h>
+#include <glm/glm.hpp>
 
 #include <string>
 #include <vector>
 
-
-
 namespace sh3_gl
 {
-
     struct program final
     {
     public:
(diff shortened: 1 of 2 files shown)


include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+        enum class Target
+        {
+            ARRAY_BUFFER    = GL_ARRAY_BUFFER,
+            ELEMENT_BUFFER  = GL_ELEMENT_ARRAY_BUFFER,
+            TEXTURE_BUFFER  = GL_TEXTURE_BUFFER
+        };
+
+        /**
+         *  Default constructor.
```
z33ky: This is not a default constructor, because it requires an argument.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+         */
+        sh3_buffer(Target type, const std::string& buffName = ""): buffType(type), name(buffName){Create();};
+
+        /**
+         *  Destructor. Ensure we have no leaks when this buffer is to be deleted by the C++ Runtime.
```
z33ky: Not necessarily by the runtime.

include/SH3/system/glbuffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+         */
+        ~sh3_buffer(){Release();};
+
+        /**
+         *  Create our buffer and register it with OpenGL.
+         */
+        void Create();
+
+        /**
+         *  Free any data related to this buffer, instead of letting the destructor do it.
```
z33ky: Is it safe to delete an OpenGL buffer multiple times?

source/SH3/system/glprogram.cpp
```diff
@@ -122,6 +122,86 @@ void program::Unbind()
...
+
+    return loc;
+}
+
+void program::SetUniformMat2(const std::string& name, const glm::mat2& mat)
+{
+    GLint uniformLoc;
+
+    uniformLoc = GetUniformLoc(name);
+    glUniformMatrix2fv(uniformLoc, 1, GL_FALSE, &mat[0][0]);
```
z33ky: These should use `glm::value_ptr(mat)` instead of `&mat[0][0]`. Same for `glm::vec*`.

include/SH3/system/glprogram.hpp
```diff
@@ -73,6 +72,87 @@ namespace sh3_gl
...
+         */
+         GLint GetUniformLoc(const std::string& name);
+
+        /**
+         *  Set the value of a 2 value matrix in this program
+         *
+         *  @param name - Name of the uniform we want to load the data into.
+         *  @param mat - The matrix we want to send to the program.
+         */
+         void SetUniformMat2(const std::string& name, const glm::mat2& mat);
```
z33ky: I think these should just use function overloading instead of different names by type.

z33ky (dismissed):

include/SH3/system/glbuffer.hpp
Quaker762: This is part of #94, as it required some of its functionality. Is there a way I can fix this??

include/SH3/system/glbuffer.hpp
Quaker762: See above

include/SH3/system/glbuffer.hpp
Quaker762: I'm not too sure, however, see #94 This causes a segmentation fault in my test program, almost like the compiler is inserting the dtor to the colour buffer, which actually releases _during_ render and causing a segmentation violation.

source/SH3/system/glprogram.cpp
Quaker762: Hmm, I didn't even know that existed (haha)

include/SH3/system/glprogram.hpp
Quaker762: Yeah I'll get on that. I wasn't too sure if it would cause confusion, so I made it explicit which uniform type was being edited in the program.

z33ky:
Ah, we'll just leave it open for now. The OpenGL VAO and VBO abstraction should land soon ;)

Quaker762:
@z33ky Guess I'll need to rebase this yes?

z33ky:
Yes.

I was thinking of having just `SetUniform(name, val)`, but it might be better to have the separation by name for `int` and `float`. The `Vec` and `Mat` postfixes then make sense for consistency.

Quaker762:
> I was thinking of having just SetUniform(name, val), but it might be better to have the separation by name for int and float. The Vec and Mat postfixes then make sense for consistency.

So we'll keep it as is then yeah?? It'll definitely help reduce type confusion, as god knows what kind of problems we might run into if we accidentally get/set the wrong type in the shader.

include/SH3/system/glprogram.hpp
```diff
@@ -17,15 +17,13 @@
 
 #include <GL/glew.h>
 #include <GL/gl.h>
+#include <glm/glm.hpp>
```
z33ky, line 20: Use `glm/fwd.hpp` here.

PR #98, opened 2017-07-14
Added way to calculate Projection Matrix

Quaker762:
Added functionality in the OpenGL context to calculate the current
Projection Matrix according to the Camera's FOV and the current Window's
dimensions, as well as being able to set the near and far values of the
current view frustum (handy for the outdoor Silent Hill level of SH3 and SH2). This may eventually need to be moved, as I don't really see a point in passing the window to any game logic (perhaps to change the Window settings in game, though this is unlike the original).

@z33ky Do you think this should be moved somewhere else?? It seems most devs put it in their 'renderer' (or in our case our context), though we could move it to the main game loop when we get further along with development.

### DIFF
--- a/include/SH3/system/glcontext.hpp
+++ b/include/SH3/system/glcontext.hpp
@@ -1,7 +1,7 @@
 /** @file
  *  Defines a logical OpenGL Context
  *
- *  @copyright 2016  Palm Studios
+ *  @copyright 2016-2017 Palm Studios
  *
  *  @date 22-12-2016
  *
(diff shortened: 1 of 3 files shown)


include/SH3/system/glcontext.hpp
```diff
@@ -15,57 +15,74 @@
...
 #include "SH3/system/sdl_destroyer.hpp"
 
 class sh3_window;
 
-class sh3_glcontext
+namespace sh3_gl
 {
-public:
-    explicit sh3_glcontext(sh3_window& hwnd);
+    class sh3_context
```
z33ky: While we're renaming, this can be just `context`.

include/SH3/system/glcontext.hpp
```diff
@@ -15,57 +15,74 @@
...
-     */
-    void PrintInfo() const;
+       /**
+        *  Print out information about the current GL Context.
+        */
+        void PrintInfo() const;
 
-private:
+       /**
+        *  Calculate the projection matrix of our
```
z33ky: "of our" ...?

include/SH3/system/glcontext.hpp
```diff
@@ -15,57 +15,74 @@
...
+        *
+        *   @param fov - The Field Of View (FOV) of our Viewing Frustum. Defined in @ref sh3_graphics::camera
```
z33ky: Add trailing dot.

include/SH3/system/window.hpp
```diff
@@ -25,10 +25,11 @@ class sh3_window
 public:
     sh3_window(int width, int height, const std::string& title);
 
-    std::unique_ptr<SDL_Window, sdl_destroyer> hwnd;
+    std::unique_ptr<SDL_Window, sdl_destroyer>  hwnd;       /**< Our window handle */
```
z33ky: An additional space sneaked in.

z33ky (changes requested):

z33ky:
> Do you think this should be moved somewhere else??

I think we should place the function in the file we need it at the moment. When we need it elsewhere, then we have additional information to decide in which module it makes sense to have the function.

include/SH3/system/glcontext.hpp
Quaker762: Whoops...

Quaker762:
Sounds good. We need a way to pass in the projection matrix to scene, as it's required to draw models and scenes (using `gl_Position = model * view * projection` etc).

include/SH3/system/glcontext.hpp
```diff
@@ -15,57 +15,76 @@
...
-    static const char* GlGetString(GLenum name);
+    private:
 
-private:
-    std::unique_ptr<flat_sdl_glcontext, sdl_destroyer> glContext;
-    std::vector<const char*> extensions;
+       /**
+        *  Get a string from the current GL Context.
+        *
+        *  @param name - Name of
```
z33ky: "Name of ..."

include/SH3/system/glcontext.hpp
```diff
@@ -15,57 +15,76 @@
...
+        *
+        *  @return char* from @c glGetString
```
z33ky: Add trailing dot.

z33ky (changes requested):

PR #99, opened 2017-07-19
Fix build with newer CMake modules

### LINKED ISSUE #92: CMakesLists.txt problems
I've switched over to Lubuntu (I gave up on Arch _literally_ needing 24/7 care to work..) and I still have problems with cmake, though they aren't as bad as Arch.

1. cmake still complains about glm not being found, however the error now gives is `file or directory //include referenced by variable GLM_INCLUDE_DIRS does not exist !`, which can be fixed by directly editing and removing a `/` from the `/../../` string in `usr/lib/cmake/glm/glmConfig.cmake`

2. I'm not sure what distro you're using @z33ky (you've mentioned it before but I've forgotten), but for me ld will fail trying to link `-lOpenGL` and `-lZLIB`, on Lubuntu they should be `-lGL` and `-lz`, which is similar to Windows (!?)

After all of these changes, I get an executable that can be run.

I'm going to start using cmake properly when I start working on this again (Code::Blocks on Windows though :P ), so I think these should be fixed up (plus I can learn a bit more about cmake)

z33ky:
This could fix #92.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -6,6 +6,17 @@ find_package(OpenGL REQUIRED)
 find_package(SDL2 REQUIRED)
 find_package(ZLIB REQUIRED)
 
+#support for older CMake versions
+if(NOT TARGET GLEW::GLEW)
+	set(GLEW::GLEW GLEW)
+endif()
+if(NOT TARGET OpenGL::GL)
+	set(OpenGL::GL OpenGL)
+endif()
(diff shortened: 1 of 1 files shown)


Quaker762 (approved):

Quaker762:
I'll give this a crack later on my Lubuntu machine. Hopefully it doesn't spit out any errors now haha.

PR #100, opened 2017-07-19
Replace explicit GLboolean checks with implicit bool conversion

z33ky:
A minor cleanup. I feel this reduces the noise a little when reading code, just like you wouldn't write `if(var == false)`. And as the commit message states, the OpenGL docs also includes examples using the implicit bool conversion, so checking against GL_FALSE is apparently not necessary for conformance.


Quaker762 (approved):

Quaker762:
Ahh okay, that makes sense. I was more or less trying to keep the OpenGL stuff standard (using the GL types) but this is probably better (and a lot cleaner)

z33ky:
Certainly. If I wouldn't have found this in the official documentation I wouldn't have proposed to change it.
And like I said I only found a couple of examples there, so it's kinda tricky to find. The documentation of GL_TRUE and GL_FALSE doesn't specify anything about their values.

Issue #80, opened 2017-02-03
Input


z33ky:
We should be able to get input at least from keyboard and mouse.
I think a stateful approach, basically maintaining a list of all pressed keys and buttons, is the best for our use case.

I'm working on a system based on SDL that also uses timestamps that can track for how long between frames a key was pressed; the time is normalized in the range [0, 1], so that it can just be multiplied for movement for instance.
It also counts how many times a key was pressed, so if something is toggled, for instance the flashlight, one can use `count % 2 == 1` to decide whether to turn it on.
I'm hoping this can improve the response in the presence of lags.

I'm ripping this code from another project of mine. [Here](https://gist.github.com/z33ky/ac5ea29aab2e41a1b7813574d511dbc5)'s some sample that shows the API's usage.
The code only checks for `State::Pressed`, but there's also the `Released`, `JustPressed` and `JustReleased` states. The `Just*` states refer to the key having been started to be pressed/released for the current update and can be used for firing a weapon, for instance.

Quaker762:
Sounds good. This should work really well for the inventory screen, which is where most of the input occurs (classic J-Horror resource management gameplay haha)

PR #101, opened 2017-07-19
Add input_system

z33ky:
Ref #80.

Please review with scrutiny. As I've stated before, I ripped the code out of another project, so check that I properly adapted the naming conventions.
I also changed some code, so check that I didn't miss updating the documentation.

Sadly the promised time-tracking is not possible with SDL, as mentioned in the documentation for `INPUT_PROVIDE_TIMING` in `input_config.hpp`.

This includes a separate test program with bindings for WASD, printing some stuff to stdout when pressing the keys. Before merging this I wish to adapt this into an automated test. We could still leave this version of the test included for documentation purposes.

### DIFF
--- a/include/SH3/system/input.hpp
+++ b/include/SH3/system/input.hpp
@@ -1,0 +1,533 @@
+/** @file
+ *  Defines the @ref sh3::system::input_system.
+ */
+#ifndef SH3_SYSTEM_INPUT_HPP_INCLUDED
+#define SH3_SYSTEM_INPUT_HPP_INCLUDED
+
+#include <bitset>
+#include <chrono>
+#include <cstdint>
+#include <ratio>
+#include <string>
+
+#include <boost/container/flat_map.hpp>
+#include <boost/operators.hpp>
+#include <boost/range/adaptor/map.hpp>
+#include <boost/range/adaptor/transformed.hpp>
+
+#include <SDL2/SDL_keyboard.h>
+#include <SDL2/SDL_mouse.h>
+#include <SDL2/SDL_timer.h>
+
+#include "input_config.hpp"
+
+struct vector2
+{
+    double x, y;
+};
+
+namespace sh3 { namespace system {
+    /**
+     *  Status of various HIDs.
+     *  
+     *  Currently only keyboard and mouse.
+     */
+    struct input_system
+    {
+    public:
+        /**
+         *  The identifier for an action.
+         *  
+         *  An @ref action is usually bound to a key or a button.
+         */
+        // Using a fixed sized structure does not seem worth it.
+        typedef std::string action;
+
+        /**
+         *  The state of an @ref action.
+         */
+        //TODO: better name?
+        struct action_state final
+        {
+        public:
+            /**
+             *  Implements SDL Ticks as std TrivialClock concept
+             */
+            struct sdl_clock final
+            {
+            public:
+                sdl_clock() = delete;
+
+                //the sdl_clock should perhaps be somewhere else
+                using duration = std::chrono::duration<Uint32, std::milli>;
+                using rep =  duration::rep;
+                using period = duration::period;
+                struct time_point : public std::chrono::time_point<sdl_clock>
+                {
+                public:
+                    /**
+                     *  Constructor.
+                     */
+                    time_point() = default;
+
+                    /**
+                     *  Constructor.
+                     *  
+                     *  @param timestamp The SDL timestamp.
+                     */
+                    explicit time_point(const Uint32 timestamp) noexcept : time_point(duration(timestamp)) {}
+
+                    using std::chrono::time_point<sdl_clock>::time_point;
+                };
+
+                static constexpr bool is_steady = true;
+
+                static time_point now() noexcept { return time_point(SDL_GetTicks()); }
+            };
+            using timestamp = sdl_clock::time_point;
+            using time_delta = sdl_clock::duration;
+
+            /**
+             *  State of an @ref action.
+             */
+            struct state final
+            {
+            public:
+                /**
+                 *  Partial @ref state representation.
+                 *  
+                 *  Used for public interface.
+                 */
+                enum simple : bool
+                {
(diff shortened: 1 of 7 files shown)


include/SH3/system/input.hpp
```diff
@@ -0,0 +1,533 @@
...
+#include <SDL2/SDL_timer.h>
+
+#include "input_config.hpp"
+
+struct vector2
+{
+    double x, y;
+};
+
+namespace sh3 { namespace system {
```
Quaker762, line 29: I'll have to go through and change the namespaces so they conform to this. At this point I'm not sure how much difference there is between SH2 and SH3, but it's probably best we prepare early.

include/SH3/system/input_config.hpp
```diff
@@ -0,0 +1,31 @@
...
+#endif
+/**
+ *  Whether the @ref sh3::system::input_system should provide counting support for inputs.
+ *  
+ *  With this enabled, the @ref sh3::system::input_system keeps track how many times an action was triggered within a frame.
+ */
+#ifdef DOXYGEN
+#define INPUT_PROVIDE_COUNT 1
+#else
+#define INPUT_PROVIDE_COUNT
```
Quaker762, line 28: Should this be `#define INPUT_PROVIDE_COUNT 0`??

include/SH3/system/input_config.hpp
```diff
@@ -0,0 +1,31 @@
...
+/**
+ *  Whether the @ref sh3::system::input_system should provide timing support for inputs.
+ *  
+ *  With this enabled, the @ref sh3::system::input_system keeps track what fraction of a frame the actions were active for.
+ *  
+ * @note Doesn't work right with SDL timestamps.
+ *       I think the event times are not when the input *happened*, but when the event-loop *received* them, which is not useful to us.
+ */
+#ifdef DOXYGEN
+#define INPUT_PROVIDE_TIMING 0
```
Quaker762, line 16: So the timestamp for each keypress is incorrect due to `SDL_PollEvent()` etc? What functionality are we sacrificing by not having this? IIRC, Silent Hill doesn't really depend on any timed input (thank God).

Also, it would seem you are correct. The timestamp occurs when the `SDL_Event` is actually handled (which is kind of annoying)

https://www.gamedev.net/forums/topic/682928-how-good-are-sdl-event-timestamps/

include/SH3/system/input.hpp
z33ky: I think it makes sense to use namespaces as we use directories.
And yes, many things we have in the SH3-directory, which is everything, are actually more generally applicable.

Moving these things about and changing to appropriate namespaces is probably a good project for someone wanting to get familiar with the source. I might have someone interested in contributing; I'll know in a week. I'll message you how it goes and if it doesn't work out either of us can do it instead.

include/SH3/system/input_config.hpp
z33ky: For the code it's just relevant which of them is `#define`d. The value doesn't matter at all.
However, you can't document `#undef`ined variables with Doxygen, so for the documentation I just `#define` them to `1` or `0` depending on whether the features are enabled or not. In the actual build, the enabled features are simply `#define`d symbols and the disabled ones are not.

include/SH3/system/input_config.hpp
z33ky: The timestamps would be used for `input_system::action_state::NormalizedTime`. Without this feature it'll return either `0` or `1`, with it, it could be any value in between, indicating for what fraction of a frame the key was pressed.
It's not a super important or even necessarily useful feature. You could let the character move for `distance * NormalizedTime(PRESSED)`. Without this feature, the character moves the whole distance, while with it, if it worked properly, it'd move corresponding to how long the button was pressed. Perhaps the only time it's even noticeable would be when the game lags. For single-player we can remedy this effect by bounding the delta-time on updates, so the game itself slows down if a desired frametime cannot be reached.

include/SH3/system/input_config.hpp
Quaker762: Ohh, okay I see. That makes more sense now.

include/SH3/system/input_config.hpp
Quaker762: > You could let the character move for distance * NormalizedTime(PRESSED)

Ahh, that's a lot clearer. I'll have to look into the movement system of both the games to see what percentage of a unit the character moves per frame (as there could be sneaking animations if you use a joystick controller). 

> For single-player we can remedy this effect by bounding the delta-time on updates, so the game itself slows down if a desired frametime cannot be reached.

Sounds good.

include/SH3/system/input.hpp
Quaker762: > I might have someone interested in contributing; I'll know in a week. I'll message you how it goes and if it doesn't work out either of us can do it instead.

Thank God, it's about time we got some more people working on this hahaha. 

> Moving these things about and changing to appropriate namespaces is probably a good project for someone wanting to get familiar with the source

Definitely, it's a little bit of a clusterfuck at the moment (though in a neat way haha :])

z33ky:
I still wanted to do the automated test thing, but no biggie, that'll just be a separate commit then.

More importantly, I forgot to tell you that this code depends on Boost (just the header-only part), which is also not added to the build system yet. Are you okay with that? I could get rid of it with some additional work.

Quaker762:
> More importantly, I forgot to tell you that this code depends on Boost (just the header-only part), which is also not added to the build system yet. Are you okay with that? I could get rid of it with some additional work.

Nah it's okay, I'll just download the boost headers and libraries. Is there any reason to _not_ use it?? The only reason I've heard is it's so big that it can bloat your binary, but that was a _loooong_ time ago.

z33ky:
Boost can be over-engineered for some use-cases, which can be noticeable mostly in compile-time. I think less so in run-time behavior, since it has high code quality and is tested in a variety of compilers and platforms.
Sometimes I look for smaller alternatives to some of Boosts libs, but generally I like their stuff.

PR #102, opened 2017-07-20
Assume ASSERTs hold in release build

z33ky:
I'm not sure what the best would be when we use neither gcc, clang or MSVC, as then we don't know how to tell the compiler that something holds true.
As mentioned in the comments, we could generate a compile-time error.
Maybe we could trigger UB, which would allow the compiler to assume the branch never happens.
For now it doesn't do anything special in that case. Maybe that's also the best way to go about it.
We could also try for a `#warning`, which will just error for compilers which do not support this non-standard directive.


Quaker762:
> We could also try for a #warning, which will just error for compilers which do not support this non-standard directive.

I like this idea as well.

z33ky:
You merged a tad early then :/ Oh well, next PR incoming :)

PR #106, opened 2017-07-28
Allow specifying buffer_object::Targets for VAO buffers

z33ky:
Note that this requires C++14 for `std::index_sequence`, which we need for unpacking the `Attributes::Targets` array. I didn't look hard for an alternative that works with C++11, hoping that this change is fine.


source/main.cpp
```diff
@@ -67,18 +86,22 @@ int main(int argc, char** argv)
...
+    using Triangle = sh3_gl::vao<TriangleAttributes>;
+
+    Triangle triVao;
+
+    triVao.Bind();
+
+    sh3_gl::buffer_object& verts = triVao[Triangle::Slot::VERTEX];
     verts.BufferData(g_vertex_buffer_data, sizeof(g_vertex_buffer_data), GL_STATIC_DRAW);
-    myVao.SetDataLocation(MyAttributes::VERTEX_SLOT, sh3_gl::vao<MyAttributes>::DataType::FLOAT, 3, 0, 0);
+    triVao.SetDataLocation(Triangle::Slot::VERTEX, Triangle::DataType::FLOAT, 3, 0, 0);
```
Quaker762, line 98: @z33ky So `Triangle::Slot::VERTEX` needs to be mapped to the same position as the one in `sh3_gl::vao_target_array<Slot> Targets` yes?

Quaker762:
> I didn't look hard for an alternative that works with C++11, hoping that this change is fine.

No worries, I'll have to do a bit of research on C++14 as I only know only the basics of 11, but I should be fine. I'll update the Code::Blocks file as well.

By the way, is your friend still going to make any changes? If so, I'll hold off on any new features (*cough*models*cough* and the camera) until he's made/moved stuff around.

source/main.cpp
z33ky: You're using words from different concepts that confuse my brain. I think the answer is "yes" though.

The `Target` for `triVao[slot]` is `TriangleAttributes::Targets[slot]`.
So if we had `Targets = { Target::ARRAY_BUFFER, Target::TEXTURE_BUFFER }`, then the vertex buffer would be `ARRAY_BUFFER` and the color buffer would be `TEXTURE_BUFFER`.

Quaker762: Ahh okay, I used the wrong terminology, it works as I thought.

Issue #95, opened 2017-06-18
Generic resource loading


z33ky:
To load resources like textures and models from the ARC files, a templated `resource` class was proposed:
```
template<typename Header> // where Header is sh3_texture_header for example
class resource
{
    ...

    Header& header() { return reinterpret_cast<Header&>(raw.front()); }

    using iterator_range = std::pair<raw::iterator, raw::iterator>; // we could use boost::iterator_range instead
    iterator_range data() { return std::make_pair(begin(raw) + header().data_offset(), end(raw)); }

    ...

    std::vector<uint8_t> raw;
};
```
A `texture` would then be a `resource<sh3_texture_header>`, which provides us access to the header by simply aliasing the bytes in `resource::raw `, as well as convenient access to the "body" (non-header bytes). Reading in the vector is then enough to load a resource from a file.

Some resource types, e.g. models, may have multiple sub-headers. This is what the `data_offset()` function of the header is for.
For headers without sub-headers, the implementation is just `return sizeof(*this);`, while for those with sub-headers, the implementation is `return sizeof(*this) + num_sub_headers * sizeof(SubHeader);` to account for them.
The sub-headers in models refer to embedded textures. These resources can also be handled with an appropriate `data_offset()` function.

Quaker762:
This is in place of `vfile` yes? So instead of having a buffer that we seek and read data from (similar to how an actual `FILE` works), we just have a `resource` with appropriate header passed in that we can more easily read bytes from the header (without a local copy in a `Load( )` function). Is there any advantage to this besides the added abstraction of the header types? If this is in place of `vfile`, would it work considering the Master File Table (arc.arc) is just a whole lot of folders and files shoved into one huge file??

EDIT: I've just reread #43 and I think I may have the wrong idea about this (it also midnight here and my reading comprehension isn't so good)

z33ky:
A `vfile` provides access to a file in the MFT as bytes, while this would interpret the bytes as a resource.
So it doesn't necessarily make `vfile` obsolete (at least in the short term), but would definitely be used in place of it.

Since the `resource` includes a local buffer (`raw`), just like the `vfile`, I don't see why it would conflict with the workings of the MFT.

Quaker762:
Okay, I think I see now. It's basically a better way to abstract the resource type and also saves a heap of code reuse (such as having heaps of calls to `file.ReadData(...)`) as the data is already there for us to access. 

[...]

PR #109, opened 2017-08-19
sh3_arc* cleanup

z33ky:
I decided the first step towards #95 is some cleanup in the arc* structs.
The rest should follow in the next days so I'd suggest you leave this PR unmerged until they are in. But you can start reviewing each commit (I'm gonna change each struct in a separate commit), hence me opening the PR already.


include/SH3/arc/mft.hpp
```diff
@@ -32,6 +34,28 @@ namespace sh3 { namespace arc {
             void operator()(gzFile file) const { gzclose(file); }
         };
 
+        /** @defgroup arc-headers arc header types
+         *  @{
+         */
+
+        /** Type check header */
+        struct Header
```
Quaker762: Should I start using Uppercase for the first letter for structs from now on??

z33ky: Ahh that's just my personal style bleeding through. I'll fix it.

Quaker762:
@z33ky Have you made the changes to this?? If so, I'll merge it :)

z33ky:
Sorry, I've been distracted last weekend by a SBC I got and have been busy otherwise this week.
I should be able to push two more commits today though; the changes are in, I just need to split them up a bit. After this there's only `vfile` left.

z33ky:
There we go. I'll squash the fixup! after review and now only `sh3_arc_vfile` is missing the refactorization treatment.

Quaker762:
Looks really good, much cleaner than what we originally had. I'm curious though, how does this line `std::string &&subarcName` work, with the double ampersand??

Apart from that, squash when you're ready :)

z33ky:
It's an rvalue-reference. Basically it's a reference where the caller promises that the value will not get used any further, which means you can "move" stuff out of it.
So for the case of the `std::string`, the copy-constructor (assuming no CoW (Copy-on-Write) semantics) needs to allocate new memory and copy the characters over, while the move-constructor just takes the existing memory from the other string and changes that string to be empty, so its destruction will not delete the memory.

z33ky:
Rebased onto master and squashed the fixup. There are two new commits on top for `vfile` with which I now consider this PR mergeable after you approve.

Quaker762 (approved):

PR #110, opened 2017-08-20
Fix sh3_arc::LoadFile

z33ky:
I tried loading the file `data/eff_tex/fire00_tr.pic`, which resulted in a 0-sized buffer.
After this change, the buffer is 266416 bytes in size, which seems more realistic. Interpreting the start of the data as a texture header tells us this is one 512x512 texture with 8bpp, which would be 262144 bytes (as is the `texSize` field), so it's that and 4272 bytes extra? `texFileSize` is just 262240. *shrug*


Quaker762:
> so it's that and 4272 bytes extra? texFileSize is just 262240. shrug

That's rather interesting, we may have to look into that a bit more if it starts giving us headaches down the track.

PR #114, opened 2017-09-24
Initial implementation of the SILENT HILL camera. Currently can perform

Quaker762:
a transformation to vertex data to project a 3D image, as well as the
position of camera (as well as its lookat() position) which is able to
be set. There is currently no collision detection/camera bounds, as we
haven't reversed that yet.

### DIFF
--- a/include/SH3/camera/camera.hpp
+++ b/include/SH3/camera/camera.hpp
@@ -1,0 +1,210 @@
+/** @file
+ *
+ *  OpenGL Version of the SILENT HILL style camera.
+ *
+ *  The camera works on a few basic principles, firstly, we consider that Heather (or James),
+ *  is the origin that the camera is <i>always</i> looking at. The camera can be in it's own position,
+ *  however for the most part (not including cutscenes), we are always looking at the protagonist.
+ *
+ *  Next, the camera seems to be in its own 'volume' (for lack of a better term), in that it <i>can't</i>
+ *  under any circumstances clip through a wall. There's a few links (albeit for different game engines) that
+ *  suggest how it is possible to do this, found here:
+ *  <a>https://www.reddit.com/r/Unity3D/comments/38qsx1/camera_collision_silent_hill_style/?st=j71bhmzc&sh=28a00dc6</a>
+ *  <a>https://answers.unrealengine.com/questions/169933/set-up-a-camera-like-silent-hill-and-resident-evil.html</a>
+ *
+ *  I googled <a href = https://www.google.com.au/search?q=silent+hill+style+camera&oq=silent+hill+style+camera&aqs=chrome..69i57j69i60j69i57j69i65j69i60l2.2672j0j7&sourceid=chrome&ie=UTF-8>this</a> term.
+ *
+ *  It would seem that camera data is stored in .cam files in the arc sections. I assume that this data is fed into the camera
+ *  when some kind of trigger (an AABB??) is triggered when Heather walks over/through it. It is currently
+ *  unknown where this data is stored, however, Mike seems to think it is in the .cld files (Collision & Lighting Data [CLD]).
+ *
+ *
+ *  There are a few modes the camera can be in:
+ *      @ref STATIC - Static camera view, usually in tight spots (the bookstore store room).
+ *      @ref FOLLOW - Follows Heather around (outside areas).
+ *  @copyright 2017 Palm Studios
+ *
+ *  @date 1-9-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#ifndef CAMERA_HPP_INCLUDED
+#define CAMERA_HPP_INCLUDED
+
+#include "glm/glm.hpp"
+#include "SH3/angle.hpp"
+
+namespace sh3{namespace camera{
+
+    /**
+     *  Camera Modes.
+     */
+    enum class MODE
+    {
+        //FOLLOW,
+        //STATIC,
+        //CYLINDER_DEBUG,
+        FIRST_PERSON
+    };
+
+    /**
+     *  Our camera in 3D space. Keeps track of position, intended position, rotation etc.
+     *
+     *  Supports clipping in a 3D scene (that is, it can collide as well as a 3D model).
+     */
+    struct Camera final
+    {
+    public:
+
+        /**
+         *  Constructor
+         *
+         *  @param pos Initial position of the camera.
+         *  @param fov Camera Horizontal field of view.
+         *  @param aspect Aspect ratio of screen.
+         *  @param zNear Near Z value for frustum culling.
+         *  @param zFar Far Z value for frustum culling.
+         */
+        Camera(const glm::vec3& pos, angle<float> fov, float aspect, float zNear, float zFar);
+
+        /**
+         *  Set the physical position of the camera.
+         *
+         *  @param pos Position vector in 3D space we want the camera at.
+         */
+        void SetPosition(const glm::vec3& pos);
+
+        /**
+         *  Look at position @ref look in 3D space.
+         *
+         *  @param look 3D vector representing the position we want the camera to look at.
+         */
+        void LookAt(const glm::vec3& look);
+
+        /**
+         *  Translates the cameras current position by adding to the position vector.
+         *
+         *  @param trans Our translation vector.
+         */
+        void Translate(const glm::vec3& trans);
+
+        /**
+         *  Translates the cameras current position by adding the camera's Front Vector,
+         *  @ref camFront, and then multiplying by a scale (which effectively chooses)
+         *  the direction.
+         *
+         *  @param factor Translation factor.
+         */
+        void Translate(const float factor);
+
+        /**
+         *  Set the mode of the camera.
+         *
+         *  @param @ref MODE we want the camera to be in.
+         */
+        void SetMode(MODE mode);
+
+        /**
+         *  Set the horizontal FOV of the camera's projection matrix.
+         *
+         *  @param fov Field of view we want the camera to be.
+         */
+        void SetFOV(angle<float> fov);
+
+        /**
+         *  Yaw the camera. There is currently no protection against gimbal lock!
+         *
+         *  @param yaw Yaw to add to the camera's current yaw.
+         */
+        void AddYaw(angle<float> yaw);
+
+        /**
+         *  Pitch the camera. There is currently no protection against gimbal lock!
+         *
+         *  @note Pitch is clamped between [-89, 89] degrees!
+         *
+         *  @param pitch Pitch to add to the camera's current pitch.
+         */
+        void AddPitch(angle<float> pitch);
+
+        /**
+         *  Calculate the projection matrix of our camera by using glm::lookat.
+         *
+         *  @return glm::mat4 that is our projection matrix. Sent to the shader as part of the MVP matrix (it is the P part).
+         */
+        const glm::mat4& GetProjectionMatrix() const;
+
+        #if 0
+        /**
+         *  Update the camera so we know where we should be each frame.
+         */
+         void Update();
+        #endif
+
+        /**
+         *  Get x co-ordinate to camera position.
+         */
+        float GetX() const;
+
+        /**
+         *  Get y co-ordinate to camera position.
+         */
+        float GetY() const;
+
+        /**
+         *  Get z co-ordinate to camera position.
+         */
+        float GetZ() const;
+
+        /**
+         *  Get the current yaw of the camera in degrees.
+         *
+         *  @return @ref angle representing the current yaw of the camera.
+         */
+        angle<float> GetYaw() const;
+
+        /**
+         *  Get the current pitch of the camera in degrees.
+         *
+         *  @return @ref angle representing the current pitch of the camera.
+         */
+        angle<float> GetPitch() const;
+
+    private:
+
+        /**
+         *  Recalculate the perspecitve matrix from updated values;
+         */
+        void RecalculatePerspective();
+
+    private:
+        glm::vec3       camPos;             /**< Position of the physical camera in 3D space. */
+        angle<float>    camFov;             /**< Camera's Field of View. */
+        float           aRatio;             /**< Aspect ratio for the camera. */
+        float           camNear;            /**< Near Z value for frustum culling. */
+        float           camFar;             /**< Far Z value for frustim culling (the reason we have fog!!) */
+        glm::vec3       camFront;           /**< Where the camera is looking. Most likely Heather. (This is actually reversed!)*/
+        glm::vec3       camUp;              /**< Up vector of the camera (not the actual world). */
+        glm::mat4       pMatrix;            /**< Our camera's Perspective matrix. */
+        //glm::mat4       vMatrix;            /**< Our camera's View Matrix. */
+        glm::vec3       camRight;           /**< Camera's Right vector. */
+
+        angle<float>    camPitch;           /**< Camera's pitch (angle on the y-axis). */
+        angle<float>    camYaw;             /**< Camera's yaw (angle on the x-axis). */
+        MODE            camMode;            /**< What mode the camera is currently in. */
+
+        // bool shouldReset;       /**< Should we refresh the perspective matrix??? */
+
+        /** Camera Physics Variables and methods */
+
+    public:
+        static constexpr double FRICTION_COEFFICIENT = 0.07;    /**< Camera Friction coefficient! TODO: Play with me! */
+
+    private:
+        // double          camVelocity;        /**< Current camera velocity */
+
+
+    };
+}}
+
+#endif // CAMERA_HPP_INCLUDED
--- a/source/SH3/camera/camera.cpp
+++ b/source/SH3/camera/camera.cpp
@@ -1,0 +1,130 @@
+/** @file
+ *
+ *  Implementation of camera.hpp
+ *
+ *  @copyright 2017 Palm Studios
+ *
+ *  @date 1-9-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include <cmath>
+#include "SH3/camera/camera.hpp"
+#include "glm/gtc/matrix_transform.hpp"
+#include <algorithm>
+#include <boost/algorithm/clamp.hpp>
+
+static constexpr glm::vec3 worldUp = glm::vec3(0, 1, 0); /**< Which axis is considered 'world up'. */
+
+sh3::camera::Camera::Camera(const glm::vec3& pos, angle<float> fov, float aspect, float zNear, float zFar)
+: camPos(pos), camFov(fov), aRatio(aspect), camNear(zNear), camFar(zFar),
+  camFront(glm::vec3(0.0f, 0.0f, -1.0f)), camUp(glm::vec3(0.0f, 1.0f, 0.0f)),
+  camPitch(angle<float>::FromDegrees(0.0f)),
+  camYaw(angle<float>::FromDegrees(-90.0f)),
+  camMode(MODE::FIRST_PERSON)
+{
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::RecalculatePerspective()
+{
+    glm::vec3 front;
+    front.x = std::cos(camPitch.AsRadians()) * std::cos(camYaw.AsRadians());
+    front.y = std::sin(camPitch.AsRadians());
+    front.z = std::cos(camPitch.AsRadians()) * std::sin(camYaw.AsRadians());
+    camFront = glm::normalize(front);
+
+    camRight = glm::normalize(glm::cross(camFront, worldUp));
+    camUp = glm::normalize(glm::cross(camRight, camFront));
+
+    //vMatrix = glm::lookAt(camPos, camPos + camFront, camUp);
+    pMatrix = glm::perspective(camFov.AsRadians(), aRatio, camNear, camFar) * glm::lookAt(camPos, camPos + camFront, camUp); // Set our perspective matrix
+}
+
+const glm::mat4& sh3::camera::Camera::GetProjectionMatrix() const
+{
+    return pMatrix;
+}
+
+void sh3::camera::Camera::SetPosition(const glm::vec3& pos)
+{
+    camPos = pos;
+}
+
+void sh3::camera::Camera::Translate(const glm::vec3& trans)
+{
+    if(camMode == MODE::FIRST_PERSON)
+    {
+        camPos += trans;
+        RecalculatePerspective();
+    }
+}
+
+void sh3::camera::Camera::Translate(const float factor)
+{
+    if(camMode == MODE::FIRST_PERSON)
+    {
+        camPos += camFront * factor;
+        RecalculatePerspective();
+    }
+}
+
+void sh3::camera::Camera::SetFOV(angle<float> fov)
+{
+    camFov = fov;
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::SetMode(sh3::camera::MODE mode)
+{
+    camMode = mode;
+}
+
+void sh3::camera::Camera::AddYaw(angle<float> yaw)
+{
+    camYaw += yaw;
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::AddPitch(angle<float> pitch)
+{
+    //FIXME: This clamp causes a weird perspective when clamped between [-90, 90], which could be due to the way we are calculating the right vector (@ref camRight)
+    camPitch = boost::algorithm::clamp(camPitch + pitch, angle<float>::FromDegrees(-89.0f), angle<float>::FromDegrees(89.0f));
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::LookAt(const glm::vec3& look)
+{
+    glm::vec3 lookDir = camPos - look;
+    camYaw += camYaw.FromRadians(-std::asin(lookDir.x / (std::sqrt((lookDir.x * lookDir.x) + (lookDir.z * lookDir.z)))));
+    camPitch += camPitch.FromRadians(-std::atan(lookDir.y/lookDir.z));
+    RecalculatePerspective();
+}
+
+float sh3::camera::Camera::GetX() const
+{
+    return camPos.x;
+}
+
+float sh3::camera::Camera::GetY() const
+{
+    return camPos.y;
+}
+
+float sh3::camera::Camera::GetZ() const
+{
+    return camPos.z;
+}
+
+angle<float> sh3::camera::Camera::GetYaw() const
+{
+    return camYaw;
+}
+
+angle<float> sh3::camera::Camera::GetPitch() const
+{
+    return camPitch;
+}
+
+
+ 
--- a/tests/camera.cpp
+++ b/tests/camera.cpp
@@ -1,0 +1,168 @@
+/** @file
+ *  Program entry point.
+ *
+ *  @copyright 2006  Palm Studios
+ *
+ *  @author Jesse Buhagiar
+ */
+
+#include "SH3/arc/mft.hpp"
+#include "SH3/system/config.hpp"
+#include "SH3/system/exit_code.hpp"
+#include "SH3/system/log.hpp"
+#include "SH3/system/window.hpp"
+#include "SH3/system/glprogram.hpp"
+#include "SH3/system/glbuffer.hpp"
+#include "SH3/system/glvertarray.hpp"
+#include "SDL2/SDL.h"
+#include "SH3/camera/camera.hpp"
+#include <cstdio>
+
+static GLfloat g_vertex_buffer_data[] = {
+   -1.0f, -1.0f, 0.0f,
+   1.0f, -1.0f, 0.0f,
+   0.0f,  1.0f, 0.0f,
+};
+
+static GLfloat glCol[] = {
+    1.0f, 0.0f, 0.0f,
+    0.0f, 1.0f, 0.0f,
+    0.0f, 0.0f, 1.0f,
+};
+
+/**
+ *  VAO Attributes for a colorful triangle.
+ */
+struct TriangleAttributes final
+{
+private:
+    using Target = sh3_gl::buffer_object::Target;
+
+    TriangleAttributes() = delete;
+
+public:
+    enum class Slot
+    {
+        VERTEX,
+        COLOR,
+        MAX
+    };
+
+    static constexpr sh3_gl::vao_target_array<Slot> Targets =
+    {{
+        Target::ARRAY_BUFFER,
+        Target::ARRAY_BUFFER
+    }};
+};
+constexpr sh3_gl::vao_target_array<TriangleAttributes::Slot> TriangleAttributes::Targets;
+
+/**
+ *  Entry point to the program.
+ *
+ *  @note Requires @c argc and @c argv to make it compatible with SDL2.
+ *
+ *  @param argc Number of arguments.
+ *  @param argv Argument vector.
+ *
+ *  @returns An @ref exit_code specifying success or failure.
+ */
+int main(int argc, char** argv)
+{
+    static_cast<void>(argc);
+    static_cast<void>(argv);
+
+
+    Log(LogLevel::INFO, "===SILENT HILL 3 REDUX===");
+    Log(LogLevel::INFO, "Copyright 2016-2017 Palm Studios\n");
+
+    sh3_window window(640, 480, "sh3redux");
+
+    bool quit = false;
+    SDL_Event e;
+
+    sh3_gl::program::load_error err;
+    sh3_gl::program test("test", err);
+
+    using Triangle = sh3_gl::vao<TriangleAttributes>;
+
+    Triangle triVao;
+
+    triVao.Bind();
+
+    sh3_gl::buffer_object& verts = triVao[Triangle::Slot::VERTEX];
+    verts.BufferData(g_vertex_buffer_data, sizeof(g_vertex_buffer_data), GL_STATIC_DRAW);
+    triVao.SetDataLocation(Triangle::Slot::VERTEX, Triangle::DataType::FLOAT, 3, 0, 0);
+
+    sh3_gl::buffer_object& col = triVao[Triangle::Slot::COLOR];
+    col.BufferData(glCol, sizeof(glCol), GL_STATIC_DRAW);
+    triVao.SetDataLocation(Triangle::Slot::COLOR, Triangle::DataType::FLOAT, 3, 0, 0);
+
+    triVao.Unbind();
+
+    sh3::camera::Camera cam(glm::vec3(0, 0, 0), angle<float>::FromDegrees(60.0f), static_cast<float>(640)/static_cast<float>(480), 0.001f, 100.0f);
+    cam.SetPosition(glm::vec3(3, 4, 3));
+    cam.LookAt(glm::vec3(0, 0, 0));
+    cam.SetMode(sh3::camera::MODE::FIRST_PERSON);
+
+    glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+
+    GLint mID = glGetUniformLocation(test.GetProgramID(), "mvp");
+    glm::mat4 model{1.0f};
+
+    bool forward = false;
+    bool backward = false;
+
+    while(!quit)
+    {
+        while(SDL_PollEvent(&e) != 0)
+        {
+            if(e.type == SDL_QUIT)
+                quit = true;
+
+            if(e.type == SDL_KEYDOWN)
+            {
+                if(e.key.keysym.sym == SDLK_w || e.key.keysym.sym == SDLK_UP)
+                    forward = true;
+
+                if(e.key.keysym.sym == SDLK_s || e.key.keysym.sym == SDLK_DOWN)
+                    backward = true;
+
+                if(e.key.keysym.sym == SDLK_ESCAPE)
+                    quit = true;
+            }
+
+            if(e.type == SDL_KEYUP)
+            {
+                if(e.key.keysym.sym == SDLK_w || e.key.keysym.sym == SDLK_UP)
+                    forward = false;
+
+                if(e.key.keysym.sym == SDLK_s || e.key.keysym.sym == SDLK_DOWN)
+                    backward = false;
+            }
+
+            if(e.type == SDL_MOUSEMOTION)
+            {
+                int deltax = e.motion.xrel;
+                int deltay = e.motion.yrel;
+
+                cam.AddPitch(angle<float>::FromDegrees(static_cast<float>(-deltay) * 0.3f));
+                cam.AddYaw(angle<float>::FromDegrees(static_cast<float>(deltax) * 0.3f));
+            }
+        }
+
+        if(forward)
+            cam.Translate(0.2f);
+        if(backward)
+            cam.Translate(-0.2f);
+
+        glClear(GL_COLOR_BUFFER_BIT);
+        test.Bind();
+        glm::mat4 mvp = cam.GetProjectionMatrix() * model;
+        glUniformMatrix4fv(mID, 1, GL_FALSE, &mvp[0][0]);
+        triVao.Draw();
+        SDL_SetRelativeMouseMode(SDL_TRUE);
+        SDL_GL_SwapWindow(window.hwnd.get());
+    }
+
+    return static_cast<int>(exit_code::SUCCESS);
+}
--- a/data/shaders/test.vert
+++ b/data/shaders/test.vert
@@ -3,8 +3,8 @@ layout(location = 0) in vec3 vertexPosition_modelspace;
 layout(location = 1) in vec3 vertexColor;
 
 out vec3 fragColor;
+uniform mat4 mvp;
 void main(){
-  gl_Position.xyz = vertexPosition_modelspace;
-  gl_Position.w = 1.0;
-  fragColor = vertexColor;
+	gl_Position = mvp * vec4(vertexPosition_modelspace,1);
+	fragColor = vertexColor;
 }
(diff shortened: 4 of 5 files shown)


Quaker762:
@z33ky We should have some 3D stuff soon! There's no way to currently control the camera (I haven't really looked at the input side of things), but its position etc can be set!

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+#define CAMERA_HPP_INCLUDED
+
+#include "glm/glm.hpp"
+
+namespace sh3{namespace camera{
+
+    /**
+     *  Camera Modes.
+     */
+    enum class CAMERA_MODE
```
z33ky: Move to `Camera::MODE`.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+         *  @arg pos Initial position of the camera.
+         *  @arg fov Camera field of view.
+         *  @arg aspect Aspect ratio of screen.
+         *  @arg zNear Near Z value for frustum culling.
+         *  @arg zFar Far Z value for frustum culling.
+         */
+        Camera(glm::vec3 pos, float fov, float aspect, float zNear, float zFar);
+
+        /**
+         *  Set the physical position of the camera.
```
z33ky: What does "physical" mean in this context?

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+        /**
+         *  Constructor
+         *
+         *  @arg pos Initial position of the camera.
+         *  @arg fov Camera field of view.
+         *  @arg aspect Aspect ratio of screen.
+         *  @arg zNear Near Z value for frustum culling.
+         *  @arg zFar Far Z value for frustum culling.
+         */
+        Camera(glm::vec3 pos, float fov, float aspect, float zNear, float zFar);
```
z33ky: The `glm::vec3` should be a `const&`.
Also for the other methods.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+         *  Set the mode of the camera.
+         *
+         *  @arg @ref CAMERA_MODE we want the camera to be in.
+         */
+        void SetMode(CAMERA_MODE mode);
+
+        /**
+         *  Recalculate the perspecitve matrix from updated values;
+         */
+        void RecalculatePerspective();
```
z33ky: `private`?

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,96 @@
...
+ *  @copyright 2017 Palm Studios
+ *
+ *  @date 1-9-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include "SH3/camera/camera.hpp"
+#include "glm/gtc/matrix_transform.hpp"
+
+static constexpr glm::vec3 camFront = glm::vec3(0.0f, 0.0f, -1.0f); /**< Camera Front. Z is INTO the screen. */
```
z33ky: Perhaps `camForward` instead?

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,96 @@
...
+sh3::camera::Camera::Camera(glm::vec3 pos, float fov, float aspect, float zNear, float zFar)
+{
+    camPos  = pos;
+    camFov  = fov;
+    aRatio  = aspect;
+    camNear = zNear;
+    camFar  = zFar;
+
+    pMatrix = glm::perspective(glm::radians(fov), aspect, zNear, zFar); // Set our perspective matrix
+    vMatrix = glm::lookAt(camPos, camPos + camFront, camUp);
```
z33ky: Use an initialization list.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+    struct Camera final
+    {
+    public:
+        Camera(){};
+
+        /**
+         *  Constructor
+         *
+         *  @arg pos Initial position of the camera.
+         *  @arg fov Camera field of view.
```
z33ky: Specify that fov is in radians.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+ *
+ *  I googled <a href = https://www.google.com.au/search?q=silent+hill+style+camera&oq=silent+hill+style+camera&aqs=chrome..69i57j69i60j69i57j69i65j69i60l2.2672j0j7&sourceid=chrome&ie=UTF-8>this</a> term.
+ *
+ *  It would seem that camera data is stored in .cam files in the arc sections. I assume that this data is fed into the camera
+ *  when some kind of trigger (an AABB??) is triggered when Heather walks over/through it. It is currently
+ *  unknown where this data is stored, however, Mike seems to think it is in the .cld files (Collision & Lighting Data [CLD]).
+ *
+ *
+ *  There are a few modes the camera can be in:
+ *      STATIC - Static camera view, usually in tight spots (the bookstore store room).
```
z33ky: @ref these modes.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+     */
+    enum class CAMERA_MODE
+    {
+        FOLLOW,
+        STATIC,
+        CYLINDER_DEBUG,
+        FIRST_PERSON
+    };
+
+    struct Camera final
```
z33ky: Class is missing documentation.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+    {
+    public:
+        Camera(){};
```
z33ky: Do we need this default ctor?

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+         *  Calculate the projection matrix of our camera by using glm::lookat.
+         *
+         *  @return glm::mat4 that is our projection matrix. Sent to the shader as part of the MVP matrix (it is the P part).
+         */
+        glm::mat4 GetProjectionMatrix() const;
+
+        /**
+         *  Update the camera so we know where we should be each frame.
+         */
+        void Update();
```
z33ky: Undefined

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+void sh3::camera::Camera::RecalculatePerspective()
+{
+    vMatrix = glm::lookAt(camPos, camPos + camFront, camUp);
+    pMatrix = glm::perspective(glm::radians(camFov), aRatio, camNear, camFar); // Set our perspective matrix
```
z33ky: We could pre-calculate `glm::radians(camFov)` in the ctor and `SetFov`.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+        glm::vec3 camPos;       /**< Position of the physical camera in 3D space. */
+        glm::vec3 camLook;      /**< Where the camera is looking. Most likely Heather. */
+
+        float aRatio;           /**< Aspect ratio for the camera. */
+        float camFov;           /**< Camera's Field of View. */
+        float camNear;          /**< Near Z value for frustum culling. */
+        float camFar;           /**< Far Z value for frustim culling (the reason we have fog!!) */
+        CAMERA_MODE camMode;    /**< What mode the camera is currently in. */
+
+        bool shouldReset;       /**< Should we refresh the perspective matrix??? */
```
z33ky: Unused.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+void sh3::camera::Camera::SetPosition(glm::vec3 pos)
+{
+    camPos = pos;
+}
+
+void sh3::camera::Camera::Translate(float x, float y, float z)
+{
+    // This is _really_ fucking messy... Please make me less hacky!!! Thanks!
```
z33ky: Since `camFront` and `camUp` are always `(0, 0, -1)` and `(0, 1, 0)`, `camFront x camUp` is always `(1, 0, 0)`, correct? As you can see you don't even need to normalize, as both operands to the product are unit vectors, so the resulting vector is as well.

Now with a positive x, camPos get's translated by `-(x, 0, 0) = (-x, 0, 0)`.
With a negative x it's `(-x, 0, 0)`, which is identical.
Why do we need the different branches then? Even an `x = 0` is handled gracefully.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+        void LookAt(glm::vec3 look);
+
+        /**
+         *  Translates the cameras current position by adding to the position vector.
+         *
+         *  @arg x Translation of n units in the x direction.
+         *  @arg y Translation of n units in the y direction.
+         *  @arg z Translation of n units in the z direction.
+         */
+        void Translate(float x, float y, float z);
```
z33ky: Why is there no `Translate(glm::vec3)`?

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+        /**
+         *  Set the mode of the camera.
+         *
+         *  @arg @ref CAMERA_MODE we want the camera to be in.
+         */
+        void SetMode(CAMERA_MODE mode);
```
z33ky: Undefined

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,169 @@
...
+#include "glm/glm.hpp"
+
+namespace sh3{namespace camera{
+
+    /**
+     *  Camera Modes.
+     */
+    enum class CAMERA_MODE
+    {
+        FOLLOW,
```
z33ky: As none of these modes are implemented yet, they shouldn't have associated values. Comment them out.

z33ky (changes requested):
`@arg` is for lists. You want `@param` to document function parameters.

I would remove the `(float, float, float)` overloads and just use `Method(glm::vec3{x, y, z})` at the call-site.

I'm thinking whether it makes sense to have strongly typed degrees & radians types. For now we can just use `float`, but I'll draft something up.
We'll also want to at least use a typedef for the `float`s, `using Length = float;` or something.

include/SH3/camera/camera.hpp
Quaker762: I like to have one in case there isn't a need for the default values, such as after a level has been loaded etc. So when everything's done, we can just feed the camera data with `SetPosition` etc

include/SH3/camera/camera.hpp
Quaker762: Whoops, I thought I was missing something...

include/SH3/camera/camera.hpp
Quaker762: Where the camera _actually_ is in a 3D space. They're not relative or anything. (1, 1, 1) would relate to the camera being at position x = 1 y = 1 z = 1 etc

z33ky: I'd prefer "absolute" then.

include/SH3/camera/camera.hpp
z33ky: Good point.

source/SH3/camera/camera.cpp
Quaker762: Interesting, I completely neglected to realise this...

Quaker762:
> I'm thinking whether it makes sense to have strongly typed degrees & radians types. For now we can just use float, but I'll draft something up.

That's a good idea, I'll review the PR now.

This should be good to go now. I'll open another PR though, as it relies (rather, the shader program) on the shader's ProgramID, which we currently can't get publicly.

Quaker762:
@z33ky eek, I think I fucked something up...

z33ky:
You should be able to resolve this by rebasing on master and when it says "Patch failed at * Add angle struct" do `git rebase --skip`.

Quaker762:
Wew, okay fixed. Don't rebase yet, as it relies on code from another branch.

z33ky (changes requested):
`angle` should be used for the method parameters as well, not just internally. This way it'll be obvious at call-site how the number is interpreted.

Quaker762:
@z33ky This should be ready to merge (totally forgot about this pr...). 

After this, I should be able to test models and see if they are loaded correctly etc. Once that done we should probably give animation a crack.

Pretty soon we're going to need to work out some framework for levels/menus etc (some kind of state manager??). I'm exactly sure how the game works out which level to load, but there's a huge function in the binary that loads every single file type (for a particular level), so I assume one of the file formats has interaction data for which level to load.

z33ky:
What's the reason for keeping the FOV parameter a `float` instead of an `angle`?
Also at least the documentation should specify that it sets the *horizontal* FOV.

Quaker762:
@z33ky I must have missed that, I could have sworn I had it as an `angle`..... I'll fix it now.

Quaker762:
@z33ky ~~This relies on code that hasn't been added yet (`program.GetID()`). I'll create another commit for it.~~

Guess who forgot about #116 

This should work as intended haha.

tests/camera.cpp
```diff
@@ -0,0 +1,177 @@
...
+    verts.BufferData(g_vertex_buffer_data, sizeof(g_vertex_buffer_data), GL_STATIC_DRAW);
+    triVao.SetDataLocation(Triangle::Slot::VERTEX, Triangle::DataType::FLOAT, 3, 0, 0);
+
+    sh3_gl::buffer_object& col = triVao[Triangle::Slot::COLOR];
+    col.BufferData(glCol, sizeof(glCol), GL_STATIC_DRAW);
+    triVao.SetDataLocation(Triangle::Slot::COLOR, Triangle::DataType::FLOAT, 3, 0, 0);
+
+    triVao.Unbind();
+
+    sh3::camera::Camera cam(glm::vec3(0,0,0), fov.FromDegrees(60), (float)640/(float)480, 0.001f, 100.0f);
```
z33ky: You can use `angle<float>::FromDegrees(60)` and remove `fov`.

tests/camera.cpp
```diff
@@ -0,0 +1,177 @@
...
+    cam.SetPosition(glm::vec3(3, 0, 4));
+    cam.LookAt(glm::vec3(0, 0, 0));
+    cam.SetMode(sh3::camera::MODE::FIRST_PERSON);
+
+    glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+
+    GLuint mID = glGetUniformLocation(test.GetID(), "mvp");
```
z33ky: Should be `GetProgramID()`.

tests/camera.cpp
```diff
@@ -0,0 +1,177 @@
...
+    glm::mat4 model = glm::mat4(1.0f);
+    float x = 0.0f;
+
+    bool forward = false;
+    bool backward = false;
+
+    angle<float> pitch;
+    angle<float> yaw;
```
z33ky: I thought the camera was broken for me, but it turns out that `yaw` is initialized with 0, but the camera is initialized with a different value.
Either the camera should get a `AddYaw` or a `GetYaw` function so we don't need to keep track of the angles ourselves (same with `pitch`).

z33ky (changes requested):
You've included some tabs in the source, while we have been consistently using spaces for indentation. Please replace the tabs by 4 spaces.

I'm getting a couple of compiler warnings (clang):
```
[1/3] Building CXX object tests/CMakeFiles/cam.dir/__/source/SH3/camera/camera.cpp.o
In file included from ../source/SH3/camera/camera.cpp:12:
../tests/../include/SH3/camera/camera.hpp:58:19: warning: extra ';' after member function definition [-Wextra-semi]
        Camera(){};
                  ^
../source/SH3/camera/camera.cpp:18:48: warning: field 'camFar' will be initialized after field 'camFront' [-Wreorder]
: camPos(pos), aRatio(aspect), camNear(zNear), camFar(zFar),
                                               ^
../source/SH3/camera/camera.cpp:19:43: warning: field 'camUp' will be initialized after field 'pMatrix' [-Wreorder]
  camFront(glm::vec3(0.0f, 0.0f, -1.0f)), camUp(glm::vec3(0.0f, 1.0f, 0.0f)),
                                          ^
../source/SH3/camera/camera.cpp:21:31: warning: field 'camPos' is uninitialized when used here [-Wuninitialized]
  vMatrix(glm::lookAt(camPos, camPos + camFront, camUp))
                              ^
../source/SH3/camera/camera.cpp:21:40: warning: field 'camFront' is uninitialized when used here [-Wuninitialized]
  vMatrix(glm::lookAt(camPos, camPos + camFront, camUp))
                                       ^
../source/SH3/camera/camera.cpp:97:40: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
        std::printf("pitch: %.2f\n", camPitch.AsDegrees());
        ~~~                          ~~~~~~~~~^~~~~~~~~~~
In file included from ../source/SH3/camera/camera.cpp:12:
../tests/../include/SH3/camera/camera.hpp:207:25: warning: private field 'camVelocity' is not used [-Wunused-private-field]
        double          camVelocity;        /**< Current camera velocity */
                        ^
7 warnings generated.
[2/3] Building CXX object tests/CMakeFiles/cam.dir/camera.cpp.o
In file included from ../tests/camera.cpp:18:
../tests/../include/SH3/camera/camera.hpp:58:19: warning: extra ';' after member function definition [-Wextra-semi]
        Camera(){};
                  ^
../tests/camera.cpp:53:9: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        Target::ARRAY_BUFFER,
        ^~~~~~~~~~~~~~~~~~~~~
        {
../tests/camera.cpp:104:68: warning: use of old-style cast [-Wold-style-cast]
    sh3::camera::Camera cam(glm::vec3(0,0,0), fov.FromDegrees(60), (float)640/(float)480, 0.001f, 100.0f);
                                                                   ^      ~~~
../tests/camera.cpp:104:79: warning: use of old-style cast [-Wold-style-cast]
    sh3::camera::Camera cam(glm::vec3(0,0,0), fov.FromDegrees(60), (float)640/(float)480, 0.001f, 100.0f);
                                                                              ^      ~~~
../tests/camera.cpp:111:18: warning: implicit conversion changes signedness: 'GLint' (aka 'int') to 'GLuint' (aka 'unsigned int') [-Wsign-conversion]
    GLuint mID = glGetUniformLocation(test.GetProgramID(), "mvp");
           ~~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/GL/glew.h:1931:43: note: expanded from macro 'glGetUniformLocation'
#define glGetUniformLocation GLEW_GET_FUN(__glewGetUniformLocation)
                                          ^
/usr/include/GL/glew.h:23656:25: note: expanded from macro 'GLEW_GET_FUN'
#define GLEW_GET_FUN(x) x
                        ^
../tests/camera.cpp:158:38: warning: implicit conversion loses floating-point precision: 'double' to 'float' [-Wconversion]
                pitch += pitch.FromDegrees(-deltay * 0.03);
                         ~~~~~             ~~~~~~~~^~~~~~
../tests/camera.cpp:161:33: warning: implicit conversion loses floating-point precision: 'double' to 'float' [-Wconversion]
                yaw += yaw.FromDegrees(deltax * 0.03);
                       ~~~             ~~~~~~~^~~~~~
../tests/camera.cpp:113:11: warning: unused variable 'x' [-Wunused-variable]
    float x = 0.0f;
          ^
```
and an additional from gcc:
```
../tests/camera.cpp:174:56: warning: conversion to ‘GLint {aka int}’ from ‘GLuint {aka unsigned int}’ may change the sign of the result [-Wsign-conversion]
         glUniformMatrix4fv(mID, 1, GL_FALSE, &mvp[0][0]);
                                                        ^
```
Please resolve those.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+void sh3::camera::Camera::Pitch(angle<float> pitch)
+{
+
+	if(pitch.AsDegrees() >= 90 || pitch.AsDegrees() <= -90)
+		return;
+
+	camPitch = pitch;
+	std::printf("pitch: %.2f\n", camPitch.AsDegrees());
```
z33ky: Remove this.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,142 @@
...
+    if(camMode == MODE::FIRST_PERSON)
+    {
+        camPos += camFront * factor;
+        RecalculatePerspective();
+    }
+}
+
+void sh3::camera::Camera::SetFOV(angle<float> fov)
+{
+    //camFov = camFov.FromDegrees(fov);
```
z33ky: Remove this.

z33ky (commented):
You might also want to rebase your branch on master. This way you get less warning-spam.

Quaker762:
> You've included some tabs in the source, while we have been consistently using spaces for indentation. Please replace the tabs by 4 spaces.

Interesting, I must have modified my vim settings somewhere... I've done a `retab` of all files with `set tabstop=4 set expandtab` and `set shiftwidth=4` which should rectify this.

> Please resolve those.

They have been fixed. I never realised the order of the initialisation list mattered in regards to variable declaration in the class' `private/public` section. Sorry :P

tests/camera.cpp
Quaker762: `cam.Pitch()` is supposed to this, but I was using the function entirely wrong...

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,212 @@
...
+     */
+    struct Camera final
+    {
+    public:
+
+        /**
+         *  Constructor
+         *
+         *  @param pos Initial position of the camera.
+         *  @param fov Camera field of view (this is in pi-radians).
```
z33ky: We no longer need to specify "pi-radians", though maybe it should specify horizontal FOV.

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+        /**
+         *  Get z co-ordinate to camera position.
+         */
+        float GetZ() const;
+
+        /**
+         *  Get the current yaw of the camera in degrees.
+         */
+        float GetYawDegrees() const;
```
z33ky: Gust return `angle` and let the caller do `AsDegrees` or `AsRadians` if it is required.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,154 @@
...
+ *
+ *  @author Jesse Buhagiar
+ */
+#include <cmath>
+#include "SH3/camera/camera.hpp"
+#include "glm/gtc/matrix_transform.hpp"
+
+static constexpr glm::vec3 worldUp        = glm::vec3(0, 1, 0);           /**< Camera Up. Y is UP. */
+
+/*
```
z33ky: Can this be removed?

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,154 @@
...
+ *
+ *  @date 1-9-2017
+ *
+ *  @author Jesse Buhagiar
+ */
+#include <cmath>
+#include "SH3/camera/camera.hpp"
+#include "glm/gtc/matrix_transform.hpp"
+
+static constexpr glm::vec3 worldUp        = glm::vec3(0, 1, 0);           /**< Camera Up. Y is UP. */
```
z33ky: What's with all those spaces?

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,154 @@
...
+void sh3::camera::Camera::Yaw(angle<float> yaw)
+{
+    camYaw += yaw;
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::Pitch(angle<float> pitch)
+{
+
+    if((camPitch.AsDegrees() + pitch.AsDegrees()) >= 90 || (camPitch.AsDegrees() + pitch.AsDegrees()) <= -90)
```
z33ky: If `camPitch` is `80` and `pitch` is `11`, this would do an early return, while it should set `camPitch` to `90` instead.
Perhaps do `camPitch = boost::algorithm::clamp(camPitch + pitch, angle<float>::FromDegrees(-90), std::max(angle<float>::FromDegrees(90));` instead.
Also prefer `x op angle<float>::FromDegrees(constant)` to `x.AsDegrees() op constant`, since with the former the conversion is done in compile-time, while for the latter it is done in run-time (unless perhaps if `-ffast-math` is set).
`angle<float>::Turn() / 4` would also be shorter than `angle<float>::FromDegrees(90)`, though I don't really have a preference to which one is used. We could also add a user-defined literal so we can do `90_float_degrees` or `90_degrees_float` or something among those lines. We can postpone this though.

tests/camera.cpp
```diff
@@ -0,0 +1,168 @@
...
+                if(e.key.keysym.sym == SDLK_s || e.key.keysym.sym == SDLK_DOWN)
+                    backward = false;
+            }
+
+            if(e.type == SDL_MOUSEMOTION)
+            {
+                int deltax = e.motion.xrel;
+                int deltay = e.motion.yrel;
+
+                cam.Pitch(angle<float>::FromDegrees(deltay * 0.3));
```
z33ky: `-deltay * 0.3f` :)

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,212 @@
...
+         *  @param fov Field of view we want the camera to be.
+         */
+        void SetFOV(angle<float> fov);
+
+        /**
+         *  Yaw the camera. There is currently no protection against gimbal lock!
+         *
+         *  @param yaw Yaw to add to the camera's current yaw.
+         */
+        void Yaw(angle<float> yaw);
```
z33ky: Maybe these should be `Add*`?
Also document that yaw is clamped.

z33ky (changes requested):
Not all warnings have been fixed. Here's the new list:
```
../source/SH3/camera/camera.cpp:38:15: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  camVelocity(0.0f)
             ~^~~~
In file included from ../source/SH3/camera/camera.cpp:12:
../tests/../include/SH3/camera/camera.hpp:206:25: warning: private field 'camVelocity' is not used [-Wunused-private-field]
        double          camVelocity;        /**< Current camera velocity */
                        ^
../tests/camera.cpp:53:9: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        Target::ARRAY_BUFFER,
        ^~~~~~~~~~~~~~~~~~~~~
        {
../tests/camera.cpp:102:81: warning: use of old-style cast [-Wold-style-cast]
    sh3::camera::Camera cam(glm::vec3(0,0,0), angle<float>::FromDegrees(60.0f), (float)640/(float)480, 0.001f, 100.0f);
                                                                                ^      ~~~
../tests/camera.cpp:102:92: warning: use of old-style cast [-Wold-style-cast]
    sh3::camera::Camera cam(glm::vec3(0,0,0), angle<float>::FromDegrees(60.0f), (float)640/(float)480, 0.001f, 100.0f);
                                                                                           ^      ~~~
../tests/camera.cpp:109:18: warning: implicit conversion changes signedness: 'GLint' (aka 'int') to 'GLuint' (aka 'unsigned int') [-Wsign-conversion]
    GLuint mID = glGetUniformLocation(test.GetProgramID(), "mvp");
           ~~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tests/camera.cpp:148:61: warning: implicit conversion loses floating-point precision: 'double' to 'float' [-Wconversion]
                cam.Pitch(angle<float>::FromDegrees(-deltay * 0.3));
                          ~~~~~                     ~~~~~~~~^~~~~
../tests/camera.cpp:149:58: warning: implicit conversion loses floating-point precision: 'double' to 'float' [-Wconversion]
                cam.Yaw(angle<float>::FromDegrees(deltax * 0.3));
                        ~~~~~                     ~~~~~~~^~~~~
```
and an additional one from gcc:
```
../tests/camera.cpp:161:56: warning: conversion to ‘GLint {aka int}’ from ‘GLuint {aka unsigned int}’ may change the sign of the result [-Wsign-conversion]
         glUniformMatrix4fv(mID, 1, GL_FALSE, &mvp[0][0]);
                                                        ^
```

z33ky:
> I never realised the order of the initialisation list mattered

It matters if the initializations have side-effects, e.g. with `my_thing::my_thing(int n): a(++n), b(++n) {}` and the creation `my_thing thing{0};`, the values of `thing.a` and `thing.b` will depend on the order in which `a` and `b` are defined in `my_thing`. If `a` comes after `b`, then `thing.a == 2 && thing.b == 1` will hold, which can be counterintuitive.

source/SH3/camera/camera.cpp
Quaker762: For the moment, this is required by RecalculatePerspective(). It tells the camera which axis is 'up' in relation to the world when it works out the view matrix. It'll probably get refactored out later on though.

source/SH3/camera/camera.cpp
Quaker762: Whoops....

source/SH3/camera/camera.cpp
z33ky: I meant the comment block that starts here.

Quaker762: LOL whoops I totally missed that one... It's gone now

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+void sh3::camera::Camera::AddYaw(angle<float> yaw)
+{
+    camYaw += yaw;
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::AddPitch(angle<float> pitch)
+{
+    camPitch = boost::algorithm::clamp(camPitch + pitch, angle<float>::FromDegrees(-89.0f), angle<float>::FromDegrees(89.0));
```
Quaker762: I was having REALLY weird perspective 'flipping' issues when clamping to [-90, 90], so 89 will do for now (FPS mode isn't really important right now anyway.)

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,208 @@
...
+         *  @param pitch Pitch to add to the camera's current pitch.
+         */
+        void AddPitch(angle<float> pitch);
+
+        /**
+         *  Calculate the projection matrix of our camera by using glm::lookat.
+         *
+         *  @return glm::mat4 that is our projection matrix. Sent to the shader as part of the MVP matrix (it is the P part).
+         */
+        glm::mat4 GetProjectionMatrix() const;
```
z33ky: I'm guessing a 'glm::mat4' is relatively expensive to copy, so perhaps return a `const glm::mat4&`?

include/SH3/camera/camera.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+        /**
```
z33ky: This comment block will be merged with the next one for Doxygen. Use `#if 0` instead and "hide" the comment block along with the method declaration. The method declaration then also doesn't need to be commented.

source/SH3/camera/camera.cpp
z33ky: Yeah, we can look at it later. Add a `FIXME` though. I think this might have something to do with calculating the right vector (`camRight = camFront x worldUp`); add this as a note to the `FIXME`.
That second `89.0` should be a `89.0f`.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "glm/gtc/matrix_transform.hpp"
+#include <algorithm>
+#include <boost/algorithm/clamp.hpp>
+
+static constexpr glm::vec3 worldUp = glm::vec3(0, 1, 0); /**< Which axis is considered 'world up'. */
+
+sh3::camera::Camera::Camera(const glm::vec3& pos, angle<float> fov, float aspect, float zNear, float zFar)
+: camPos(pos), camFov(fov), aRatio(aspect), camNear(zNear), camFar(zFar),
+  camFront(glm::vec3(0.0f, 0.0f, -1.0f)), camUp(glm::vec3(0.0f, 1.0f, 0.0f)),
+  pMatrix(glm::perspective(0.0f, aspect, zNear, zFar)),
```
z33ky: We don't need to initialize `pMatrix` and `vMatrix` since that is done in `RecalculatePerspective`.

tests/camera.cpp
```diff
@@ -0,0 +1,168 @@
...
+                if(e.key.keysym.sym == SDLK_s || e.key.keysym.sym == SDLK_DOWN)
+                    backward = false;
+            }
+
+            if(e.type == SDL_MOUSEMOTION)
+            {
+                int deltax = e.motion.xrel;
+                int deltay = e.motion.yrel;
+
+                cam.Pitch(angle<float>::FromDegrees(static_cast<float>(-deltay) * 0.3f));
```
z33ky: Update these function calls.

tests/camera.cpp
```diff
@@ -0,0 +1,168 @@
...
+public:
+    enum class Slot
+    {
+        VERTEX,
+        COLOR,
+        MAX
+    };
+
+    static constexpr sh3_gl::vao_target_array<Slot> Targets =
+    {
```
z33ky: I'm still getting the `-Wmissing-braces` warning on clang. Add another set of curly braces.

The warnings seems spurious: https://bugs.llvm.org/show_bug.cgi?id=21629
For now this is an easy fix though.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+void sh3::camera::Camera::AddPitch(angle<float> pitch)
+{
+    camPitch = boost::algorithm::clamp(camPitch + pitch, angle<float>::FromDegrees(-89.0f), angle<float>::FromDegrees(89.0));
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::LookAt(const glm::vec3& look)
+{
+    glm::vec3 lookDir = camPos - look;;
+    camYaw += camYaw.FromRadians(-asinf(lookDir.x / (sqrtf((lookDir.x * lookDir.x) + (lookDir.z * lookDir.z)))));
```
z33ky: Prefer `std::asin`, `std::sqrt` and `std::atan`.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+  camYaw(angle<float>::FromDegrees(0.0f)),
+  camMode(MODE::FIRST_PERSON)
+{
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::RecalculatePerspective()
+{
+    glm::vec3 front;
+    front.x = cos(camPitch.AsRadians()) * cos(camYaw.AsRadians());
```
z33ky: Prefer `std::cos` and `std::sin`.

tests/camera.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    sh3::camera::Camera cam(glm::vec3(0,0,0), angle<float>::FromDegrees(60.0f), static_cast<float>(640)/static_cast<float>(480), 0.001f, 100.0f);
+    cam.SetPosition(glm::vec3(3, 0, 4));
+    cam.LookAt(glm::vec3(1, 0, 0));
+    cam.SetMode(sh3::camera::MODE::FIRST_PERSON);
+
+    glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+
+    GLint mID = glGetUniformLocation(test.GetProgramID(), "mvp");
+    glm::mat4 model = glm::mat4(1.0f);
```
z33ky: This could be shortened to `glm::mat4 model{1.0f};`.

z33ky (changes requested):

tests/camera.cpp
Quaker762: Hmmm, that's a very strange warning to get. I've added another set of braces, hopefully that suppresses it.

include/SH3/camera/camera.hpp
Quaker762: This causes a segmentation fault (which I assume is caused by line 160 in tests/camera.cpp). Perhaps OpenGL is modifying it in the glUniformMatrix4f function? I'll add a FIXME for now (so we can work on some other stuff and get some 3D going :P )

z33ky: Ah I've assumed this would just return pMatrix. My mistake.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+    camRight = glm::normalize(glm::cross(camFront, worldUp));
+    camUp = glm::normalize(glm::cross(camRight, camFront));
+
+    vMatrix = glm::lookAt(camPos, camPos + camFront, camUp);
+    pMatrix = glm::perspective(camFov.AsRadians(), aRatio, camNear, camFar); // Set our perspective matrix
+}
+
+glm::mat4 sh3::camera::Camera::GetProjectionMatrix() const
+{
+    return pMatrix * glm::lookAt(camPos, camPos + camFront, camUp);
```
z33ky: Any reason this isn't `pMatrix * vMatrix`?
Also, can't we already do the multiplication in `RecalculatePerspective()` and store it in `pMatrix`? If so, then this method can return a `const glm::mat4&` fine.

z33ky (commented):
Hmm, the camera in the test program should initially look at the triangle, right? It doesn't seem to do that for me anymore.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+    glm::vec3 front;
+    front.x = std::cos(camPitch.AsRadians()) * std::cos(camYaw.AsRadians());
+    front.y = std::sin(camPitch.AsRadians());
+    front.z = std::cos(camPitch.AsRadians()) * std::sin(camYaw.AsRadians());
+    camFront = glm::normalize(front);
+
+    camRight = glm::normalize(glm::cross(camFront, worldUp));
+    camUp = glm::normalize(glm::cross(camRight, camFront));
+
+    vMatrix = glm::lookAt(camPos, camPos + camFront, camUp);
```
z33ky: vMatrix is only used here. Can it be removed? Or if `pMatrix` stores the product as suggested below, `vMatrix` doesn't need to remain a member of `Camera`.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+void sh3::camera::Camera::AddYaw(angle<float> yaw)
+{
+    camYaw += yaw;
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::AddPitch(angle<float> pitch)
+{
+    //FIXME: This clamp causes a weird perspective when clamped between [-90, 90], which could be due to the way we are
```
z33ky: I think you accidentially a few words.

Quaker762: Shit I did... I also forgot what I was going to write whoops HAHA.

source/SH3/camera/camera.cpp
Quaker762: Ahh very true. I've stored the product of the two matrices in pmatrix. I'm merely going to comment out vmatrix for now, as I have a feeling we may encounter a time where we'll need access to the view matrix/send it to the shader (though for the most part, the view and model matrix are actually combined, the view-model matrix.)

Quaker762:
> Hmm, the camera in the test program should initially look at the triangle, right? It doesn't seem to do that for me anymore.

I may or may not have been setting the pitch to -90f in the initialiser list when it should have actually been yaw... Oops... 

I'll have to go over the math again to make it a bit more robust, but for now it works :)

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+void sh3::camera::Camera::LookAt(const glm::vec3& look)
+{
+    glm::vec3 lookDir = camPos - look;;
+    camYaw += camYaw.FromRadians(-std::asin(lookDir.x / (std::sqrt((lookDir.x * lookDir.x) + (lookDir.z * lookDir.z)))));
+    camPitch += camPitch.FromRadians(-std::atan(lookDir.y/lookDir.z));
+    RecalculatePerspective();
+
+    std::printf("lookdir: x %.2f y %.2f z %.2f\n", lookDir.x, lookDir.y, lookDir.z);
+    std::printf("camYaw: %.2f\ncamPitch: %.2f\n", camYaw.AsDegrees(), camPitch.AsDegrees());
```
z33ky: Remove.

z33ky (changes requested):
I can't move anymore in the camera test.

source/SH3/camera/camera.cpp
```diff
@@ -0,0 +1,133 @@
...
+void sh3::camera::Camera::AddPitch(angle<float> pitch)
+{
+    //FIXME: This clamp causes a weird perspective when clamped between [-90, 90], which could be due to the way we are calculating the right vector (@ref camRight)
+    camPitch = boost::algorithm::clamp(camPitch + pitch, angle<float>::FromDegrees(-89.0f), angle<float>::FromDegrees(89.0f));
+    RecalculatePerspective();
+}
+
+void sh3::camera::Camera::LookAt(const glm::vec3& look)
+{
+    glm::vec3 lookDir = camPos - look;;
```
z33ky: Double semicolon

Quaker762:
> I can't move anymore in the camera test.

That's really strange, I still can... Do you mean you have no translation or rotation of the camera??

z33ky:
Yeah but I just noticed I was using the shader from the master branch in my build directory. Whoops, my bad. It's working fine with the correct shader :)

z33ky (approved):

PR #115, opened 2017-10-01
Add Angle struct

z33ky:
This will make it explicit when we expect to deal with angles in degrees or radians.
The internal representation is in radians, since maths functions usually expect this and thus we hopefully need less conversions.

I'm not certain if `Angle::Zero()` and `Angle::Turn()` are useful.
Because equality-comparing floating-point values is tricky, I've opted into only providing an `IsExactlyZero` method. If we ever need to check the equality of angles in the future we can make up our minds on how to implement that.

As we stumble upon them we can add overloads for certain functions, e.g. `std::min` or `std::sin`, so we can just say `std::min(angleA, angleB)` instead of `Angle::FromRadians(angleA.AsRadians(), angleB.AsRadians())`.

### DIFF
--- a/include/SH3/angle.hpp
+++ b/include/SH3/angle.hpp
@@ -1,0 +1,107 @@
+/** @file
+ *  Definition of @ref angle.
+ */
+#ifndef SH3_ANGLE_HPP_INCLUDED
+#define SH3_ANGLE_HPP_INCLUDED
+
+#include <cmath>
+
+#include <boost/math/constants/constants.hpp>
+#include <boost/operators.hpp>
+
+/**
+ *  An angle.
+ *  
+ *  @tparam T The datatype for the internal representation.
+ *  
+ *  The angle is stored in radians.
+ */
+template<typename T>
+struct angle final : public boost::additive<angle<T>>, public boost::less_than_comparable<angle<T>>, public boost::multiplicative<angle<T>, T>
+{
+public:
+    constexpr angle() = default;
+    constexpr angle(const angle&) = default;
+
+    /**
+     *  Create an 0-@ref angle.
+     */
+    static constexpr angle Zero() { return FromRadians(0); }
+    /**
+     *  Create an @ref angle representing a full turn.
+     *  
+     *  @note The returned @ref angle is not normalized, as this would just be equal to @ref Zero otherwise.
+     */
+    static constexpr angle Turn() { return FromRadians(boost::math::constants::two_pi<T>()); }
+
+    /**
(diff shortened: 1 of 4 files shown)


Quaker762:
Looks good. 

Did you want me to add this into the camera (in regards to the FOV etc) after we merge this?

z33ky:
Yes.

Quaker762:
@z33ky I can't seem to compile with this. I'm getting a whole lot of 

`undefined reference to `angle<float>::angle(float)'|`

`undefined reference to `angle<double>::angle(double)'|`

The file has been added to project and is being compiled.

z33ky:
Added angle.cpp to the Code::Blocks project.

Have you added any code using `angle`?

Quaker762:
It still doesn't seem to be working. Hrrm

> Have you added any code using angle?

Nope, this is just a straight recompile with the angle class added. Perhaps there's a bug (_again_.....) with my version of g++ in regards to using `extern` with templates??

Here's the build log:

https://pastebin.com/Drz7A81c

*EDIT:* Just downloaded and installed MinGW with g++ 7.1.0, it's still throwing the same linker error.

z33ky:
What's the output of `nm obj\Debug\i686\source\SH3\angle.o|c++filt`?

z33ky:
This seems to be a known issue: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57728
I'm guessing that the empty body has the same effect as `= default;` here.
Try removing the `extern template` lines from `angle.hpp`. If that works I'll check the compiler version to conditionally use this feature or not. It's not necessary, just a compile time improvement.

Quaker762:
> Try removing the extern template lines from angle.hpp. If that works I'll check the compiler version to conditionally use this feature or not. It's not necessary, just a compile time improvement.

This seems to have fixed it. The version of mingw g++ I have is 7.1.0 (not sure what the latest is on Linux). I should have a Linux box soon though.

include/SH3/angle.hpp
```diff
@@ -0,0 +1,103 @@
...
+    constexpr angle(T rad): radians{rad} {}
+
+private:
+    T radians;
+};
+
+using float_angle = angle<float>;
+using double_angle = angle<double>;
+
+extern template struct angle<float>;
```
Quaker762: These need to be removed to prevent the gcc template compiler bug

Quaker762 (changes requested):

z33ky:
Alright, try this.

PR #116, opened 2017-11-14
Added getter for glprogram::programID

Quaker762:
Added a way to get the programID from a shader program instance.

### DIFF
--- a/include/SH3/system/glprogram.hpp
+++ b/include/SH3/system/glprogram.hpp
@@ -73,6 +73,15 @@ namespace sh3_gl
          */
         void Unbind();
 
+        /**
+         *  Gets the program ID for this shader.
+         *  
+	 	 *	Allows us to access the instance of this shader outside of the class so that we can send/receive data on the GPU.  
+         *
+         *  return @ref programID The ID of this shader program.
+         */
+        GLuint GetProgramID() const {return programID;}
+
     private:
         GLuint programID;           /**< The ID of this program. */
         GLuint vertShader;          /**< The ID of the vertex shader. */


Quaker762:
This should be merged before #114

include/SH3/system/glprogram.hpp
```diff
@@ -73,6 +73,14 @@ namespace sh3_gl
          */
         void Unbind();
 
+        /**
+         *  Gets the OpenGL generated Program ID from instance of this (shader) class so that
+         *  we can send/get data from outside of it.
+         *
+         *  return @ref programID The ID of this shader program.
+         */
+        inline GLuint GetProgramID(){return programID;}
```
z33ky: `inline` gets implied, so we don't need to specify that.
This function can be `const` though, which cannot be inferred by the compiler.

include/SH3/system/glprogram.hpp
```diff
@@ -73,6 +73,14 @@ namespace sh3_gl
          */
         void Unbind();
 
+        /**
+         *  Gets the OpenGL generated Program ID from instance of this (shader) class so that
```
z33ky: The first sentence is a short description. I'd write something like "Gets the Program ID.". Then have an empty line, followed by the long description, if any is necessary.

z33ky (changes requested):

include/SH3/system/glprogram.hpp
Quaker762: Oh I see, so the compiler is smart enough to know that it should be inlined?? I've been reading a bit of old C++ lately so excuse the verbose syntax.

z33ky: Methods that are defined in the class definition are treated as specified with inline.

But the inline specifier is mostly just required to specify appropriate linkage. Compilers nowadays use heuristics based on the body and caller locations to determine when to inline functions. Having the definition available when it's called is essential for inlining though, unless you're compiling with LTO.

Issue #111, opened 2017-08-24
Useless warnings from gcc -Weffc++


z33ky:
We get some annoying warnings from `-Weffc++` about initializing members via the member initialization list, even though their default constructor is perfectly fine. See [GCC Bug 81431](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81431) for an upstream bug report, the referenced [GCC Bug 16166](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=16166) is also relevant.
We should try to replace this warning with other flags to report things we **do** want to hear about. At worst we could just get rid of the flag and live with less warnings.

Quaker762:
Okay, so I should update the Code::Blocks project with it disabled?? I'll see what other warnings I can access.

z33ky:
Right now we're getting warning-spammed anyways, so for the time being I don't think it does actual damage.
I believe most warnings are reasonably fixable in code. The member-initialization warnings of `-Weffc++` look 100% spurious though, but there are other warnings enabled with that same flag that are useful.

Like the OP states, I'm proposing we look how to best *replace* this flag with some others, so we can get rid of this annoying warning, while keeping the useful ones. If we cannot find good replacements, then I believe it is better to just disable it and accept less compiler-assistance.

Quaker762:
Alright then, that makes a lot more sense, it is rather annoying have five gazillion warnings spat out by the compiler.

Also, I'm thinking of moving constructors to the source files instead so it's a bit cleaner, but I'll open another issue for this.

z33ky:
I still get spurious `-Weffc++` warnings; I don't think this is "Done".

PR #117, opened 2017-11-23
Silence most warnings

z33ky:
With this PR, only `-Weffc++` warnings remain for gcc (see #111) and spurious `-Wdocumentation` warnings (because it doesn't understand `@see @ref Thing`).


Quaker762:
This fixes most of the warnings except the ones that are GCC bugs as pointed out in #111 yes?

z33ky:
All of the warnings, except for the `-Weffc++` (gcc) and `-Wdocumentation` (clang) ones.
I don't think the `-Weffc++` warnings we get are bugs, they are just warnings about things that I don't think warrant changes. As stated in #111 though, this flag also warns about things that I think do warrant changes.
The `-Wdocumentation` thing is a bug, because `@see @ref Thing` is certainly valid.

PR #118, opened 2017-12-22
Fix subarc load

z33ky:
Oops. I guess we need a better testing discipline.


Quaker762:
The whole format KONAMI used is very _very_ broken imho. It doesn't help they didn't bother deleting certain files that are actually 0 bytes/duplicated when they shipped the game out... Really just a ragtag group of programmers I suppose..

Quaker762 (approved):

PR #119, opened 2018-01-11
Fixed 8-bit textures not displaying

Quaker762:
There was a problem with the texture type in texture.cpp for 8-bit paletted textures. This has now been rectified.

### DIFF
--- a/source/SH3/graphics/texture.cpp
+++ b/source/SH3/graphics/texture.cpp
@@ -160,7 +160,7 @@ void sh3_texture::Load(sh3::arc::mft& mft, const std::string& filename)
     if(header.bpp == PixelFormat::PALETTE)
     {
         palette_info         pal_header;
-        std::vector<bgra>    palette;    // Palette Data (I think this is BGRA)
+        std::vector<rgba>    palette;    // Palette Data (I think this is BGRA)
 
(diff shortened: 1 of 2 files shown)


Quaker762:
There might be a problem with the 16-bit textures, we'll need to test them before pull.

z33ky:
I couldn't find a 16-bit texture in the trial version, though I can confirm this fixes paletted textures.

z33ky:
Actually, I think the proper fix is making the palette be a `std::vector` of `rgba` instead of `bgra`

Quaker762:
> Actually, I think the proper fix is making the palette be a std::vector of rgba instead of bgra

How do you mean? According to the debugger I was using it was caused by an invalid OpenGL source texture format. Are you referring to 16-bit textures?

z33ky:
Well, `type` should still be changed to `GL_UNSIGNED_BYTE`, but I think `srcFormat` should stay `GL_RGB`, since we're decoding the texture that way (see line 307-309, 320-322). The actual issue is that the pallette should be `rgba` I think; it just works with the current change because two wrongs make a right in this case.

Quaker762:
@z33ky  Ahhh I see what you mean, I was under the assumption that pallete was a `bgra`. I've fixed it up.

PR #123, opened 2018-01-30
Use error struct for sh3::arc::{mft,subarc}::LoadFile

z33ky:
We're currently not doing much with the error, but that can be changed easily.
I'm also doing more error checking than before.

### DIFF
--- a/source/SH3/arc/subarc.cpp
+++ b/source/SH3/arc/subarc.cpp
@@ -28,6 +28,8 @@ namespace {
         std::uint32_t numFiles;          /**< Number of files located in this sub .arc */
         std::uint32_t dataPointer;       /**< Pointer to the beginning of the data section */
         std::uint32_t unused;            /**< Unused DWORD */
+
+        static constexpr std::uint32_t expectedMagic = 0x20030507; /**< Expected value of @ref magic. */
     };
 
     /**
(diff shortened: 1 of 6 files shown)


source/SH3/arc/subarc.cpp
```diff
@@ -28,6 +28,8 @@ namespace {
         std::uint32_t numFiles;          /**< Number of files located in this sub .arc */
         std::uint32_t dataPointer;       /**< Pointer to the beginning of the data section */
         std::uint32_t unused;            /**< Unused DWORD */
+
+        static constexpr std::uint32_t expectedMagic = 0x20030507; /**< Expected value of @ref magic. */
```
Quaker762, line 32: Should this be `ALL_CAPS` for a static constant??

z33ky: It seems we're a little bit inconsistent. See `arcFileNotFound` or our `sh3_gl::vao_target_array`s for instance. We should perhaps try to define a `.clang-format` and run a pass on our code.

Quaker762: Hmm I see, I'm surprised that one slipped through as well (maybe when we refactored it).

Can we define our own formatting style with .clang-format (so that the codebase still follows the Allman style)?

PR #124, opened 2018-02-13
Added OpenGL error message debug callback

Quaker762:
Created a type definition for all OpenGL functions that are callback functions (__glCallback)
as well as enabling OpenGL's debug callback routine (instead of us having to call
glGetError constantly, which is deprecated at this point).

### DIFF
--- a/source/SH3/system/glcontext.cpp
+++ b/source/SH3/system/glcontext.cpp
@@ -17,10 +17,110 @@
 #include "SH3/system/glcontext.hpp"
 #include "SH3/system/log.hpp"
 #include "SH3/system/window.hpp"
+#include "SH3/system/assert.hpp"
 #include <GL/glew.h>
 #include <GL/glu.h>
 #include <GL/glext.h>
 #include <glm/gtc/matrix_transform.hpp>
+
+/**
+ *  OpenGL error callback function. 
+ *
+ *  Used in conjunction with glDebugMessageCallback.
+ *
+ *  @arg source     The Source of this error.
+ *  @arg type       The type of error this is.
+ *  @arg id         Error ID.
+ *  @arg severity   How severe this error is. If it's fatal, terminate the program with @ref die().
+ *  @arg length     Length of the message log.
+ *  @arg message    Error message passed to us from OpenGL.
+ *  @arg data       User Param data.
+ */
+static void GLAPIENTRY debugCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar* message, const GLvoid* data)
+{   
+    LogLevel    level;
+    std::string errMsg;
+    
+    static_cast<void>(id);
+    static_cast<void>(length);
+    
+    ASSERT(!data);
+    errMsg = "";
+
+    // Get the source of the error message
+    switch(source)
+    {
+    case GL_DEBUG_SOURCE_API: // Error caused by the OpenGL API
+        errMsg += "API ";
+        break;
+    case GL_DEBUG_SOURCE_WINDOW_SYSTEM:
+        errMsg += "WINDOW SYSTEM ";
+        break;
+    case GL_DEBUG_SOURCE_SHADER_COMPILER:
+        errMsg += "SHADER COMPILER ";
+        break;
+    case GL_DEBUG_SOURCE_APPLICATION:
+        errMsg += "APPLICATION ";
+        break;
+    case GL_DEBUG_SOURCE_OTHER:
+        errMsg += "OTHER ";
+        break;
+    default:
+        errMsg += "UNKNOWN ";
+        break;
+    }
+
+    // Get the type of error this is
+    switch(type)
+    {
+    case GL_DEBUG_TYPE_ERROR:
+        errMsg += "(ERROR): ";
+        break;
+    case GL_DEBUG_TYPE_DEPRECATED_BEHAVIOR:
+        errMsg += "(DEPRECATED): ";
+        break;
+    case GL_DEBUG_TYPE_UNDEFINED_BEHAVIOR:
+        errMsg += "(UNDEFINED BEHAVIOR): ";
+        break;
+    case GL_DEBUG_TYPE_PORTABILITY:
+        errMsg += "PORTABILITY! ";
+        break;
+    case GL_DEBUG_TYPE_PERFORMANCE:
+        errMsg += "(PERFORMANCE): ";
+        break;
+    case GL_DEBUG_TYPE_OTHER:
+        errMsg += "(OTHER): ";
+        break;
+    case GL_DEBUG_TYPE_MARKER:
+        errMsg += "(MARKER): ";
+        break;
+    default:
+        errMsg += "(UNKNOWN): ";
+        break;
+    }
+
+    switch(severity)
+    {
+        case GL_DEBUG_SEVERITY_HIGH:
+            level = LogLevel::FATAL;
+            break;
+        case GL_DEBUG_SEVERITY_MEDIUM:
+            level = LogLevel::ERROR;
+            break;
+        case GL_DEBUG_SEVERITY_LOW:
+            level = LogLevel::WARN;
+            break;
+        case GL_DEBUG_SEVERITY_NOTIFICATION:
+            level = LogLevel::INFO;
+            break;
+        default:
+            level = LogLevel::INFO;
+            break;
+    }
+
+    errMsg += message;
+    Log(level, errMsg.c_str());
+}
 
 using namespace sh3_gl;
 
@@ -28,6 +128,7 @@ context::context(sh3_window& hwnd)
 {
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_FLAGS, SDL_GL_CONTEXT_FORWARD_COMPATIBLE_FLAG); // Put the context into 'forwrad compatible' mode, meaning no deprecated functionality will be allowed AT ALL!
 
     glContext.reset(SDL_GL_CreateContext(hwnd.hwnd.get()));
 
@@ -45,7 +146,19 @@ context::context(sh3_window& hwnd)
     SDL_GL_SetAttribute(SDL_GL_BUFFER_SIZE, 32);
 
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
-}
+
+    if(GLEW_KHR_debug)
+    {
+        glEnable(GL_DEBUG_OUTPUT);
+        glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
+        glDebugMessageCallback(debugCallback, nullptr);
+        glDebugMessageControl(GL_DONT_CARE, GL_DONT_CARE, GL_DONT_CARE, 0, nullptr, GL_TRUE);
+    }
+    else
+    {
+        Log(LogLevel::WARN, "GL_KHR_debug not available on this OpenGL context!");
+    }
+}   
 
 const char* context::GetVendor() const
 {


include/SH3/glcallback.hpp
```diff
@@ -0,0 +1,21 @@
...
+ *
+ *  @author Jesse Buhagiar
+ */
+#ifndef SH3_GL_CALLBACK_HPP_INCLUDED
+#define SH3_GL_CALLBACK_HPP_INCLUDED
+
+// We use C linkage on this function typedef so we don't get any C++ name decoration (as OpenGL is a C library)
+extern "C"
+{
+    typedef void __glCallback;
```
z33ky: This is not a function though, you're just defining `__glCallback` as `void`?

source/SH3/system/glcontext.cpp
```diff
@@ -17,11 +17,109 @@
...
+ *
+ *  @arg source     The Source of this error
+ *  @arg type       The type of error this is
+ *  @arg id         Error ID
+ *  @arg severity   How severe this error is. If it's fatal, terminate the program with @ref die()
+ *  @arg length     Length of the message log
+ *  @arg message    Error message passed to us from OpenGL
+ *  @arg data       User Param data
+ */
+__glCallback debugCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *data)
```
z33ky: This can be static. And while you probably intended to have this function as `extern "C"`, the `APIENTRY`-macro is probably what we want to use here.

source/SH3/system/glcontext.cpp
```diff
@@ -17,11 +17,109 @@
...
 #include "SH3/system/log.hpp"
 #include "SH3/system/window.hpp"
+#include "SH3/glcallback.hpp"
 #include <GL/glew.h>
 #include <GL/glu.h>
 #include <GL/glext.h>
 #include <glm/gtc/matrix_transform.hpp>
 
+/**
+ *  OpenGL error callback function. Used in conjunction with glDebugMessageCallback
```
z33ky: End the documentation lines with a dot.

Add an empty line (just ` *  `) before the second sentence to make it part of only the long description.

source/SH3/system/glcontext.cpp
```diff
@@ -17,11 +17,109 @@
...
+            level = LogLevel::INFO;
+            break;
+        default:
+            level = LogLevel::INFO;
+            break;
+    }
+
+    errMsg += message;
+
+    if(severity == GL_DEBUG_SEVERITY_HIGH)
```
z33ky: I like always having braces. I think we've also always used braces on statements elsewhere.

source/SH3/system/glcontext.cpp
```diff
@@ -45,6 +143,14 @@ context::context(sh3_window& hwnd)
...
 
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
+
+    // Enable debug output (may work for non-debug contexts, depending on OpenGL implementation)
+    // Kinda risky because it __mayyy__ not work on all systems/graphics drivers, but GL 3.3 gives
+    // us the most compatability.
+    glEnable(GL_DEBUG_OUTPUT);
+    glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
+    glDebugMessageCallback(debugCallback, nullptr);
+    glDebugMessageControl(GL_DONT_CARE, GL_DONT_CARE, GL_DONT_CARE, 0, NULL, true);
```
z33ky: Use `nullptr` in place of `NULL` and `GL_TRUE` in place of `true`.

source/SH3/system/glcontext.cpp
```diff
@@ -17,11 +17,109 @@
...
+ *  @arg type       The type of error this is
+ *  @arg id         Error ID
+ *  @arg severity   How severe this error is. If it's fatal, terminate the program with @ref die()
+ *  @arg length     Length of the message log
+ *  @arg message    Error message passed to us from OpenGL
+ *  @arg data       User Param data
+ */
+__glCallback debugCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *data)
+{   
+    LogLevel    level;
```
z33ky: We could `ASSERT(!data)`.

z33ky (changes requested):
Please silence these warnings:
```
source/SH3/system/glcontext.cpp:37:63: warning: unused parameter ‘id’ [-Wunused-parameter]
 __glCallback debugCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *data)
                                                               ^~
source/SH3/system/glcontext.cpp:37:92: warning: unused parameter ‘length’ [-Wunused-parameter]
 __glCallback debugCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *data)
                                                                                            ^~~~~~
source/SH3/system/glcontext.cpp:37:135: warning: unused parameter ‘data’ [-Wunused-parameter]
 __glCallback debugCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *data)
```

Use `static_cast<void>(param)` if you want to explicitly ignore parameters.

include/SH3/glcallback.hpp
Quaker762: Oops. I was under the assumption the compiler would automatically insert `extern "C"` along with the typedef function.

source/SH3/system/glcontext.cpp
Quaker762: Yeah same here, I wasn't sure if I should do it for consistency. I think I need for fix up a few files, though if use a formatter it should sort itself out.

source/SH3/system/glcontext.cpp
Quaker762: > the APIENTRY-macro is probably what we want to use here.

The few tutorials I've seen have used this as well (though on my system it's `GLAPIENTRY`). Would you prefer having a custom `extern "C"` type we can use across files or are you okay with functions being defined as `static extern "C" void GLAPIENTRY` (I did this mostly for readability)

Quaker762: > And while you probably intended to have this function as extern "C", the APIENTRY-macro is probably what we want to use here.

`APIENTRY` (or `GLAPIENTRY`) expands to __stdcall on my system. Is there a particular purpose to this (or why it's necessary for this OpenGL function??)

source/SH3/system/glcontext.cpp
```diff
@@ -17,16 +17,122 @@
...
+
+    Log(level, errMsg.c_str());
+}
+
 using namespace sh3_gl;
 
 context::context(sh3_window& hwnd)
 {
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 4);
```
Quaker762: @z33ky I've changed the version of OpenGL we're using to 4.3 in this one as well (it's required for debugging and much better for compatibility). I don't feel the need for this to be in a separate PR, but I'll change it if you'd like :)

include/SH3/glcallback.hpp
z33ky: This is not a function typedef though. You're just typedeffing `void`, which is not a function, but a "nothing"-type.
If you wanted a typedef for a `debugCallback`-like function you could do `extern "C" typedef void(*__glCallback)(GLenum source, GLenum type, ...);` or (C++11) `using __glCallback = void(*)(GLenum Source, GLenum type, ...)`. There's also such a thing as function references in C++.
Also note that identifiers starting with a double-underscore are reserved.

source/SH3/system/glcontext.cpp
z33ky: It seems I was looking at a tutorial for C and Windows, because we need both `extern "C"` and `GL_APIENTRY`. On my system `APIENTRY` and `GL_APIENTRY` are both defined to be the same thing, though the name `GL_APIENTRY` gives more context anyways.
`__stdcall` specifies a certain way to pass arguments and the return value.

source/SH3/system/glcontext.cpp
z33ky: It'd be awesome if we could target OpenGL ES 2.0 as a "minimum", which is supported on the Raspberry Pi and OpenPandora and probably a couple of other mobile devices.
Wikipedia says it's more or less OpenGL 2.0 without the fixed-function pipeline. Using a lower version core profile should enable us to have less issues if we or someone else ever tries to port our renderer to OpenGL ES 2.0. We could check for the `KHR_debug` extension instead (see `SDL_GL_ExtensionSupported` and `SDL_GL_GetProcAddress`) and just not install the callback if it doesn't exist (and log that).

include/SH3/glcallback.hpp
Quaker762: > If you wanted a typedef for a debugCallback-like function you could do extern "C" typedef void(*__glCallback)(GLenum source, GLenum type, ...); or (C++11) using __glCallback = void(*)(GLenum Source, GLenum type, ...).

Ohhh okay that makes a lot more sense now. It's been a bit of a while since I've dealt with function pointers...

> Also note that identifiers starting with a double-underscore are reserved.

~~By the compiler??~~ Ahh I see now, I've been looking at too much embedded code... Probably not a good idea to pollute that namespace..

source/SH3/system/glcontext.cpp
Quaker762: >It'd be awesome if we could target OpenGL ES 2.0 as a "minimum", which is supported on the Raspberry Pi and OpenPandora and probably a couple of other mobile devices.

Would this be best to have some kind of preprocessor #define switch??

> Wikipedia says it's more or less OpenGL 2.0 without the fixed-function pipeline. Using a lower version core profile should enable us to have less issues if we or someone else ever tries to port our renderer to OpenGL ES 2.0. We could check for the KHR_debug extension instead (see SDL_GL_ExtensionSupported and SDL_GL_GetProcAddress) and just not install the callback if it doesn't exist (and log that).

Should I do this in another PR? I'll get on it ASAP.

> Using a lower version core profile should enable us to have less issues if we or someone else ever tries to port our renderer to OpenGL ES 2.0.

From the looks of it, trying to create an OpenGL 2.0 context on my PC (GTX470 Arch Linux) just defaults back to an OpenGL 4.6 context.

z33ky: > Would this be best to have some kind of preprocessor #define switch?

Either that or even a runtime mechanism, like function pointers that either point to an OpenGL or an OpenGL ES implementation.
For now I'd just stick with desktop OpenGL until we get stuff drawn and we have an idea of how we want to design our renderer.

> Should I do this in another PR? I'll get on it ASAP.

I'd stick with a OpenGL 3.x ~core~ forward compatibility profile for now and just use the extension mechanism for this feature. This way we also don't use inefficient, deprecated stuff from before the "modern OpenGL".
When we get to needing more stuff from OpenGL 4.x we can reevaluate whether we want to keep using extensions or switch to 4.x proper.

> From the looks of it, trying to create an OpenGL 2.0 context on my PC (GTX470 Arch Linux) just defaults back to an OpenGL 4.6 context.

I just read that OpenGL vendors may return a compatible profile. Since you can use anything of OpenGL 2.0 with a 4.6 compatibility profile this is legitimate. Sadly this means we can unintentionally use features beyond what our specified version provides. Sigh.

source/SH3/system/glcontext.cpp
```diff
@@ -45,7 +152,22 @@ context::context(sh3_window& hwnd)
     SDL_GL_SetAttribute(SDL_GL_BUFFER_SIZE, 32);
 
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
-}
+
+    // Enable debug output (may work for non-debug contexts, depending on OpenGL implementation)
+    // Kinda risky because it __mayyy__ not work on all systems/graphics drivers, but GL 3.3 gives
+    // us the most compatability.
```
z33ky: It's not really risky when we check for availability, is it?

source/SH3/system/glcontext.cpp
```diff
@@ -45,7 +152,22 @@ context::context(sh3_window& hwnd)
...
+    if(SDL_GL_ExtensionSupported("GL_KHR_debug"))
+    {
+        glEnable(GL_DEBUG_OUTPUT);
+        glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
+        glDebugMessageCallback(debugCallback, nullptr);
+        glDebugMessageControl(GL_DONT_CARE, GL_DONT_CARE, GL_DONT_CARE, 0, nullptr, GL_TRUE);
```
z33ky, line 155: We need to dynamically load the address using `SDL_GL_GetProcAddress`, otherwise the startup of our application will fail on systems where this extension is not supported.
We can additionally check for "ARB_debug_output", which functions are named differently but have an identical signature so it should work the same.
If you want to go the extra mile you can also try "AMD_debug_output", though there the functions have a differing signature so you need to change a bit more.

z33ky (changes requested):

source/SH3/system/glcontext.cpp
Quaker762: > We need to dynamically load the address using SDL_GL_GetProcAddress, otherwise the startup of our application will fail on systems where this extension is not supported.

I thought GLEW did that for us?

> We can additionally check for "ARB_debug_output", which functions are named differently but have an identical signature so it should work the same.
If you want to go the extra mile you can also try "AMD_debug_output", though there the functions have a differing signature so you need to change a bit more.

Okay I'll get on adding that just to make it extra water-tight.

source/SH3/system/glcontext.cpp
Quaker762: Whoops, I hate forgetting to remove outdated comments lol :P

source/SH3/system/glcontext.cpp
z33ky: > I thought GLEW did that for us?

Oh, you're right. You can also shorten the check to `if(GLEW_KHR_debug)` with GLEW :)

z33ky:
Looks good, but running the tex-test results in
```[fatal] API (ERROR): GL_INVALID_ENUM error generated. Cannot enable <cap> in the current profile.```.

Quaker762:
This is due to #125. It should be all fixed.

z33ky:
Oh yea, that. Makes sense :)

I wonder though if we should exit or just warn about it considering it "worked" before. I can't remember if I vocalized my thought before, but for prototyping I think it's a good idea to allow using unportable hacks without any friction.

Quaker762:
> Oh yea, that. Makes sense :)

Dw the amount of times this tripped me up when I was testing....

> I wonder though if we should exit or just warn about it considering it worked before.

It's a tad dirty, but perhaps a `#ifdef DEBUG` macro or something?? Or just log for debug and release builds?

z33ky:
I've thought about it and I think we want to only log for any build.
In release builds I think it's better to try and keep running with visual defects than outright aborting.
In debug builds we don't really gain anything from aborting either. When we get a sort of HUD on the screen we should just write these warnings there.
And we can eventually implement a test harness that also parses logs after execution.

z33ky (approved):

PR #125, opened 2018-02-13
Fixed 'tex.cpp'

Quaker762:
Removed calls to glEnable() and glDisable() as they were causing a fatal GL error (as they are deprecated) which closed the program.


Quaker762:
@z33ky Wow, turns out enabling that debug messaging is _REALLY_ effective... It caught onto the fact that `glEnable` with `GL_TEXTURE_2D` is deprecated.

z33ky:
Hmmm what's with the commented `glDisable(GL_{BLEND,LIGHTING})`? Is this something we will later need to explicitly disable here?

Quaker762:
> Hmmm what's with the commented glDisable(GL_{BLEND,LIGHTING})? Is this something we will later need to explicitly disable here?

Oops.. I pushed the wrong version. These enums are deprecated for `glEnable` so I don't think we'll be needing them any time soon, they just result in a `GL_SEVERE` warning and cause us to exit.

PR #126, opened 2018-02-14
Moved all pixel types to 'color.hpp'

Quaker762:
Considering these types need to be used elsewhere besides textures (such as vertex data) these types have been moved to their own header under 'SH3/types/'

### DIFF
--- a/include/SH3/types/color.hpp
+++ b/include/SH3/types/color.hpp
@@ -1,0 +1,68 @@
+/** @file
+ *  Macros and structures related to pixel information.
+ *
+ *  @copyright 2016  Palm Studios
+ *
+ *  @date 14-12-2016
+ *
+ *  @author Jesse Buhagiar
+ */
+#ifndef COLOR_HPP_INCLUDED
+#define COLOR_HPP_INCLUDED
+
+#include <GL/gl.h>
+#include <cstdint>
+#include <cstddef>
+
+/** @defgroup graphics-types Graphics Types
+ *  @{
+*/
+
+/**
+ *  RGBA pixel
+ */
+struct rgba
+{
+    std::uint8_t r, g, b, a;
+};
+
+/*
+ *  RGBA16 pixel
+ */
+struct rgba16
+{
+    std::uint8_t r : 5;
+    std::uint8_t g : 5;
+    std::uint8_t b : 5;
+    std::uint8_t a : 1;
+};
+
+/**
+ *  RGB24 pixel
+ */
+struct rgb
+{
+    std::uint8_t r, g, b;
+};
+
+/**
+ * BGRA pixel (swizzled RGBA)
+ */
+struct bgra
+{
+    std::uint8_t b, g, r, a;
+};
+
+/**
+ *  BGR24 pixel (swizzled RGB24)
+ */
+struct bgr
+{
+    std::uint8_t b, g, r;
+};
+
+static_assert(offsetof(struct rgb, b) == 2* sizeof(std::uint8_t), "struct has been padded on this compiler! (mingw/win32 use '-mno-ms-bitfields!')");
+
+/**@}*/
+
+#endif // COLOR_HPP_INCLUDED
--- a/source/SH3/graphics/texture.cpp
+++ b/source/SH3/graphics/texture.cpp
@@ -13,6 +13,7 @@
 #include <SH3/system/log.hpp>
 #include <SH3/arc/mft.hpp>
 #include <SH3/arc/vfile.hpp>
+#include <SH3/types/color.hpp>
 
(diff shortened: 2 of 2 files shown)


z33ky:
Apart from deserialization, do we explicitly want there to be no padding?

Quaker762:
> Apart from deserialization, do we explicitly want there to be no padding?

At the moment, not that I'm aware of. I usually pack my structs regardless (mostly because I deal with binary files). Is there any reason (besides natural alignment, obviously) that we shouldn't be??

z33ky:
The compiler usually inserts padding so that memory is aligned and access is efficient. That padding comes with a price in memory of course, but I think by default we should just rely on the compiler.

For loading the textures from the subarcs we need the packed layout of course. For that file we should define separate PackedRGB-like types.

Edit: We probably only need `rgb` and `rgba` in `color.hpp` then I guess.

Quaker762:
> For loading the textures from the subarcs we need the packed layout of course. For that file we should define separate PackedRGB-like types.

So should I put those in two seperate headers? Or just do a `#pragma` after the non-packed types??

> Edit: We probably only need rgb and rgba in color.hpp then I guess.

Hmm, I'm not sure about this. KONAMI used different types for different things (each pixel is just a type defined for each texture type). We'd have to manually swizzle stuff then, which I'm not really a fan of.

z33ky:
> So should I put those in two seperate headers? Or just do a #pragma after the non-packed types??

If we need the packed types somewhere else I'd put them in the same header.
Otherwise I'd leave them in `texture.cpp`.

> We'd have to manually swizzle stuff then, which I'm not really a fan of.

I meant only for `color.hpp`. They are useful in `texture.cpp`, but outside of it I don't expect that we will be working with anything other than `rgb(a)`.

source/SH3/graphics/texture.cpp
```diff
@@ -23,12 +23,21 @@
...
 
-#pragma pack(push, 1)
+/** @defgroup graphics-types Graphics Types
+ *  @{
+*/
+
+/**
+ *  RGBA pixel
+ */
 struct rgba
```
z33ky: These types should be wrapped in an anonymous namespace to avoid collision with the types in `color.hpp`.
I'm uncertain about the name resolution then though. I think with just `rgba` you'd get the packed version and with `::rgba` you can get the unpacked one.
Still it might be better to call them `packed_*` or something to avoid confusion.

include/SH3/types/color.hpp
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * BGRA pixel (swizzled RGBA)
+ */
+struct bgra
+{
+    std::uint8_t b, g, r, a;
+};
+
+/**
+ *  BGR24 pixel (swizzled BGRA)
```
z33ky: s/BGRA/RGB/

source/SH3/graphics/texture.cpp
```diff
@@ -23,12 +23,21 @@
 
 using namespace sh3_graphics;
 
-#pragma pack(push, 1)
```
z33ky, line 26: Actually, thinking about it, will the structures have any padding at all? I know that `char` is required to have an alignment of 1, I'm guessing that `std::uint8_t` should be the same. We could just `static_assert` that in the `color.hpp` header and be done with it.

z33ky (changes requested):

include/SH3/types/color.hpp
Quaker762: Whoops

source/SH3/graphics/texture.cpp
Quaker762: Yeah I think that would look a lot nicer than `::rgba` imho.

source/SH3/graphics/texture.cpp
Quaker762:
~~~Hmm, I neglected to think of that actually. I'll try it out and push it up if it works. Packing structs is a bit of a nasty habit I have from uni sorry.~~~

You were right (yet again). The alignment is 1 for char so no packing is required.

Quaker762:
> I meant only for color.hpp. They are useful in texture.cpp, but outside of it I don't expect that we will be working with anything other than rgb(a).

I think we should keep them all in there until we do some more digging through the games' assets. The amount of times they switch (even mid-format...) is absolutely mind boggling. Levels seem to have embedded textures that are both rgb/bgr.

z33ky:
Hmmh I guess it could be the case that they use strange formats for other information such as fog color or lighting. Fair enough.
Nice `static_assert` btw :)

z33ky (approved):

PR #128, opened 2018-02-16
Fix sh3::arc::mft::LoadFile

z33ky:
How many times have I fixed and subsequently broken file loading now?


Quaker762 (approved):

Quaker762:
> How many times have I fixed and subsequently broken file loading now?

About 4 times if my count is correct HAHAHA.

PR #129, opened 2018-02-19
Change 'Open()' to public

Quaker762:
Moved the visibility of Open() to public. Certain objects in the future (maps etc) will need to have files in which the filename is determined at runtime, without forcing us to determine it at compile time.

@z33ky I had sh3 open in a debugger all day today and realised having `Open()` private is a _horrible_ idea.. I'm not sure how it'll interfere with your resource manager, but I really think we'll need raw file handles for a lot of things, considering how the actual engine treats everything.

### DIFF
--- a/include/SH3/arc/vfile.hpp
+++ b/include/SH3/arc/vfile.hpp
@@ -50,6 +50,7 @@ namespace sh3 { namespace arc {
         };
 
 
+        vfile(){}
         vfile(mft& mft, const std::string& filename): fpos(0), fname(filename)
         {Open(mft, filename);}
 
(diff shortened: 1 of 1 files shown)


include/SH3/arc/vfile.hpp
```diff
@@ -50,6 +50,7 @@ namespace sh3 { namespace arc {
         };
 
 
+        vfile() : fpos(0), fname(""){}
```
z33ky: Integers are initialized to 0 by default. `std::string` is initialized to the empty string by default as well.

z33ky (changes requested):
This will not have any effect on my resource loady thingy.

Quaker762:
> This will not have any effect on my resource loady thingy.

I moreso meant that the engine seems to ID every file and correlates that with an offset that is then seeked in the raw file handle. Also some files are structured _very_ strangely (as you can see in Mike's code). Though I suppose for some `scene` class we could have a bunch of the templated headers (I haven't seen the code so I don't quite know how it'll work haha :) )

include/SH3/arc/vfile.hpp
Quaker762: You learn something new every day hahaha.

z33ky (approved):

PR #130, opened 2018-02-19
Tools

Quaker762:
@z33ky I'm going to merge all of the stuff I was working on w.r.t `sh3tools` here as it'll be much easier to keep track of everything. Also note that these are supposed to be _really_ quick and dirty, so code quality really shouldn't matter too much (as long as they work properly, which they do :P). Hopefully we can figure out a few of the file formats with these. Atm they rely on #129.

I'm also going to work on a tool to repack `.arc` sections so that we can modify files to see what happens. If that fails, it's possible to use sh2 (not 3) to do this, as the files are already exposed. For example, you can copy `.anm` files to say, the toilet scene, and James will end up flying around the room because the animations are all wrong (I'll take a video of it, it's hilarious).

If you want these to be a single commit, let me know and I'll squash 'em.


Quaker762:
Oops, I just realised sublime 3 has used tabs instead of spaces...

Quaker762:
This is a separate set of programs for another repo, not this one.

PR #132, opened 2018-02-25
Fixed Building on Windows (finally)

Quaker762:
It is now actually possible to compile on Windows using cmake instead of having to rely on an IDE environment. An icon for the test programs was also added to differentiate it from the actual (sh3r.exe) program.


Quaker762:
It was an absolute pain in the arse to get this to work correctly... Luckily it should work 100%, though it relies on a path passed in from `build.bat`, which is the only way I could think to do this without any centralized package location.

CMakeLists.txt
```diff
@@ -20,7 +24,7 @@ endif()
 
 include(CheckCXXCompilerFlag)
 
-if(NOT CMAKE_BUILD_TYPE)
+if(NOT CMAKE_BUIiLD_TYPE)
```
z33ky, line 27: This change can't be right.

CMakeLists.txt
```diff
@@ -125,6 +129,31 @@ elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
 	include_directories(SYSTEM "include/support/osx")
 endif()
 
+# If we're on windows, do a check for what compiler we're using (so we can link the appropriate libraries)
+if(WIN32)
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") # We're dealing with MinGW etc
+        set(Boost_USE_STATIC_LIBS ON)
+        set(BOOST_INCLUDEDIR ${LIB_ROOT_DIR}/boost/boost)
```
z33ky, line 136: `LIB_ROOT_DIR` should be a documented CMake variable that appears in the configurator (.e.g `cmake-gui` or `ccmake`).

CMakeLists.txt
```diff
@@ -125,6 +129,31 @@ elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
...
+        # Do a check for system bitness
+        if(AMD64)
+            set(BOOST_LIBRARYDIR ${LIB_ROOT_DIR}/boost/lib/amd64/debug)
+        else()
+            set(BOOST_LIBRARYDIR ${LIB_ROOT_DIR}/boost/lib/x86/debug)
+        endif()
+
+        set(GLEW_LIBRARY ${LIB_ROOT_DIR}/GLEW/lib/libglew.a)
+        set(SDL2_LIBRARY ${LIB_ROOT_DIR}/SDL2/lib)
+        set(ZLIB_LIBRARY ${LIB_ROOT_DIR}/zlib/lib/zlib.a)
```
z33ky, line 147: Wouldn't it suffice to set `GLEW_ROOT_DIR` & friends so that the libraries' CMake-configuration files search the path?
At this point `LIB_ROOT_DIR` might become the same as `CMAKE_PREFIX_PATH`. Perhaps try setting this to your "lib"-directory on Windows.

CMakeLists.txt
```diff
@@ -125,6 +129,31 @@ elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
...
+
+        set(GLEW_LIBRARIES ${LIB_ROOT_DIR}/GLEW/lib/libglew.a)
+        set(SDL2_INCLUDE_DIRS ${LIB_ROOT_DIR}/SDL2/include/SDL2) # CMake doesn't automatically pass this like the others???
+        set(CMAKE_EXE_LINKER_FLAGS "-lmingw32")
```
z33ky, line 151: Does GCC on Windows imply MinGW? What about Cygwin?

CMakeLists.txt
```diff
@@ -125,6 +129,31 @@ elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
...
+        else()
+            set(BOOST_LIBRARYDIR ${LIB_ROOT_DIR}/boost/lib/x86/debug)
+        endif()
+
+        set(GLEW_LIBRARY ${LIB_ROOT_DIR}/GLEW/lib/libglew.a)
+        set(SDL2_LIBRARY ${LIB_ROOT_DIR}/SDL2/lib)
+        set(ZLIB_LIBRARY ${LIB_ROOT_DIR}/zlib/lib/zlib.a)
+
+        set(GLEW_LIBRARIES ${LIB_ROOT_DIR}/GLEW/lib/libglew.a)
+        set(SDL2_INCLUDE_DIRS ${LIB_ROOT_DIR}/SDL2/include/SDL2) # CMake doesn't automatically pass this like the others???
```
z33ky, line 150: I think you might be referring to the `SDL2_INCLUDE_DIR` in your batch file, which might just be missing an `S`.

z33ky (changes requested):

z33ky:
Also I've used tabs in the CMake files.

CMakeLists.txt
Quaker762: I was definitely using VIM or something when I did this hahaha.

CMakeLists.txt
Quaker762: I've passed it in with -D in the .bat file, does that not create the variable? It needs to be a defined CMake one? I apologise for this, I couldn't find much info on it online.

CMakeLists.txt
Quaker762: > At this point LIB_ROOT_DIR might become the same as CMAKE_PREFIX_PATH. Perhaps try setting this to your "lib"-directory on Windows.

I think this might be the correct way to go.

CMakeLists.txt
Quaker762: I assume so, but I'll need to double check. Considering they're both built off of GCC.

Quaker762:
> Also I've used tabs in the CMake files.

Goddamn.. I'll get on this in the next few days. I've got 3 assignments due this week. Kill me.

CMakeLists.txt
z33ky: It'll work, yeah, but when you define it with `set` ([doc](https://cmake.org/cmake/help/v3.11/command/set.html#set-cache-entry)) you can document it and it will show up in the interactive CMake dialogs.

Quaker762:
This is 110% deprecated now :3

PR #139, opened 2019-01-08
Singleton Template class

Quaker762:
Basic templated singleton template class. This is used in instances
where we need one and ONLY one instance of a class, such as the engine
itself, or it's managers. Due to the nature of the original SH3 engine,
I'm going out on a limb (and especially from my research) and assuming we'regoing to need a fair few "global" variables such as these to satisfy the original data structures.

It's currently not thread safe. If two seperate execution threads
attempt to access the instance of a class inheriting this, it will
result in undefined behaviour. This could probably be fixed by using
`std::unique_ptr<T>` or `std::shared_ptr<T>`, however I want to keep it
simple unless we actually need more than one thread going at a time.
From memory, audio in the original game is on a seperate thread,
probably due to memory constraints of ~2003. We could probably just load
all the audio into memory anyways (via some kind of singleton 'manager'
or lookup table) and be done with it anyways, but I digress.

### DIFF
--- a/include/SH3/common/singleton.hpp
+++ b/include/SH3/common/singleton.hpp
@@ -1,0 +1,87 @@
+/** @file
+ *  Singleton template class. Any object that <i>must</i> have only one instance (such as the game
+ *  engine itself, or it's specific managers [i.e scene, texture, shader]) should inherit this class.
+ *
+ *  @copyright 2016-2019 Palm Studios
+ *
+ *  @date 8-1-2019
+ *
+ *  @author Jesse Buhagiar
+ */
+
+#ifndef SINGLETON_HPP
+#define SINGLETON_HPP
+
+namespace sh3
+{
+
+template<typename T>
+class CSingleton
+{
+public:
+    typedef T SingletonType; /**< Base type of this singleton */
+
+public:
+    /**
+     * Copy constructor
+     *
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    CSingleton(const CSingleton& singleton) = delete;
+
+    /**
+     * Move constructor
+     *
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    CSingleton(const CSingleton&& singleton) = delete;
+
+    /**
+     * Assignment operator.
+     *
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    T& operator=(CSingleton& singleton) = delete;
+
+    /**
+     * Move assignment operator
+     *
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    T& operator=(CSingleton&& singleton) = delete;
+
+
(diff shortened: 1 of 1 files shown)


Quaker762:
It seems my IDE has used tabs instead of spaces... I'll fix that up shortly.

z33ky:
Do we need the Destroy-method? If not we can easily make this thread-safe by using a static local non-pointer variable. static variables are guaranteed to be initialized by at most one thread.
Methods defined in class-definitions are implicitly `inline`, thus the qualifier is not needed.

Quaker762:
> Do we need the Destroy-method?

No we don't! I honestly doubt we'd need that much fine grained control.

`§6.7 [stmt.dcl] p4 If control enters the declaration concurrently while the variable is being initialized, the concurrent execution shall wait for completion of the initialization.`

I was unaware of this. I've fixed it up now. I should probably give the standard a read some time.

include/SH3/common/singleton.hpp
```diff
@@ -0,0 +1,72 @@
...
+public:
+    typedef T SingletonType; /**< Base type of this singleton */
+
+public:
+    /**
+     * Copy constructor
+     *
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    CSingleton(const CSingleton& singleton) = delete;
```
z33ky, line 30: I think that moving a singleton also doesn't make sense.

include/SH3/common/singleton.hpp
```diff
@@ -0,0 +1,72 @@
...
+    static T& Instance(void)
+    {
+        /**
+         * According to the C++ standard:
+         *
+         *
+         * �6.7 [stmt.dcl] p4 If control enters the declaration concurrently while the variable is being
+         * initialized, the concurrent execution shall wait for completion of the initialization.
+         *
+         * Obtaining the singleton instance in this way will be threadsafe
```
z33ky, line 70: The commit message should be updated, too.

z33ky (changes requested):

Quaker762:
That should do it. Let me know if I've forgotten anything else.

include/SH3/common/singleton.hpp
```diff
@@ -0,0 +1,79 @@
...
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    CSingleton(const CSingleton& singleton) = delete;
+
+    /**
+     * Move constructor
+     *
+     * @note This is deleted. We do NOT want to copy a singleton under any circumstance
+     */
+    CSingleton(const CSingleton&& singleton) = delete;
```
z33ky, line 37: const rvalue-references are rarely, if ever, a thing. In any case, the move constructor takes a non-const rvalue-reference.
There exists also a move-assignment operator which you can delete. Same thing with non-constness there.

z33ky (changes requested):

z33ky (approved):

PR #141, opened 2019-01-28
Complete overhaul of build system

Quaker762:
Build system has been overhauled, especially for systems that do NOT
have a package manager installed (i.e Windows). The `libs` folder
contains the most up to date versions of libraries we are using, which
are known to be WORKING. This way (at least for Windows builds), we
control versioning of our dependencies much better instead of having the
user have to fetch them. The premake5 file also takes into account which
system the user is on, meaning we can link linux versions of the
libraries, as WELL as windows versions.

Premake is also much less of a fuck around, and the simple `.lua` script
is very simple to modify and understand.

This is only temporary until @z33ky can start working on a meson build
system. Until then, this should keep Windows and Linux users satisfied
and straight into programming without having to fuck around with finding
dependencies online.


Quaker762:
Going to speedy merge this now until we have a better build system :) This will work for anyone using mingw-w64 8.1.0 rev0

PR #149, opened 2019-03-27
Engine modifications

Quaker762:
Final push to get the basic engine up and running. And the result is *drumroll*

https://youtu.be/_MkQlJyQOPE

@z33ky It actually fucking does something... It's been a long road, and hopefully we can keep the ball rolling!


Quaker762:
Paging doctor @z33ky Uh-oh, I broke git again......

Quaker762:
@z33ky I totally just trashed the commit log here, but it's rather inconsequential considering how experimental everything is atm (so I doubt we'll need to rewind/revert etc).

